PR #19, opened 2021-07-07
Add Iterators to Custom Objects

jeffrey-hui:
The Python response objects that just wrap around a list object (BestChoices, Embeddings, Similarities) should be iterable.

For example, Embeddings returned from the `Embed` endpoint can be iterated through by using: 
```python
# embeddings = co.embed(model="baseline-shrimp", texts=["co:here", "cohere"])

for em in embeddings.embeddings:
      print(em)
```
Users should expect to be able to do this instead:
```python
for em in embeddings:
      print(em)
```


setup.py
```diff
@@ -5,14 +5,17 @@
 
 setuptools.setup(
     name="cohere",
-    version="0.0.2",
+    version="0.0.7",
```
jimwu6, line 8: 0.0.9 (I forgot to update to 0.0.8)

jimwu6 (approved):
Added comments to separate stuff in sanity-test.py, so take a look

LGTM otherwise

PR #43, opened 2022-01-31
Batch Request For Embeds

OrenLeung:
# What it does
- [x] Add non-parallel Batch Embeds
- [x] Parallel Batch Embeds through python `Threading Workers`  or `concurrent.futures.ThreadPoolExecutor` (still researching to figure out which is better) 

## Design Decision
- chose thread executor over multi processing executor as it is IO-bound but CPU-bound so most of the time the threads are waiting for a network response
- the response must be in order (test added)

## Future PR
- [ ] add default num_workers depend on user's # of cores #46 
- [ ] exponential decay retrying #47 

## Tests
- [x] added unit tests for multiple batches
- [x] test that it is in order

## Basic Performance Test for 25k texts

- @jalammar reported it takes 5 hours for non-parallel

- batching resulted in 3mins & 30 secs

```python
letters = string.ascii_lowercase

def random_word():
    return ''.join(random.choice(letters) for i in range(10))

def random_sentence(num):
    sentence = ""

    for i in range(num):
        sentence += random_word()

    return sentence

def list_of_list(num):
    arr = []

    for i in range(num):
        arr.append(random_sentence(50))
    
    return arr

co = cohere.Client(
    API_KEY)

prediction = co.embed(
    model='small',
    texts=list_of_list(25000))

print(len(prediction.embeddings))
```

### DIFF
--- a/cohere/__init__.py
+++ b/cohere/__init__.py
@@ -3,6 +3,7 @@ from .error import CohereError
 
 COHERE_API_URL = 'https://api.cohere.ai'
 COHERE_VERSION = '2021-11-08'
+COHERE_BATCH_SIZE = 5
 GENERATE_URL = 'generate'
 EMBED_URL = 'embed'
 CHOOSE_BEST_URL = 'choose-best'
--- a/cohere/__init__.py
+++ b/cohere/__init__.py
@@ -3,7 +3,7 @@ from .error import CohereError
 
 COHERE_API_URL = 'https://api.cohere.ai'
 COHERE_VERSION = '2021-11-08'
-COHERE_BATCH_SIZE = 5
+COHERE_EMBED_BATCH_SIZE = 5
 GENERATE_URL = 'generate'
 EMBED_URL = 'embed'
 CHOOSE_BEST_URL = 'choose-best'
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -2,8 +2,12 @@ import json
 from typing import List, Any
 from urllib.parse import urljoin
 
+import math
+
 import requests
 from requests import Response
+
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import cohere
 from cohere.best_choices import BestChoices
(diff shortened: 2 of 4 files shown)


cohere/__init__.py
```diff
@@ -3,6 +3,7 @@
 
 COHERE_API_URL = 'https://api.cohere.ai'
 COHERE_VERSION = '2021-11-08'
+COHERE_BATCH_SIZE = 5
```
jimwu6, line 6: perhaps we should say `COHERE_EMBED_BATCH_SIZE`?

cohere/client.py
```diff
@@ -90,21 +94,33 @@ def generate(
...
+                json_body = json.dumps({
+                    'texts': text,
+                    'truncate': truncate,
+                })
+                request_futures.append(executor.submit(
+                    self.__request, json_body, cohere.EMBED_URL, model))
+
+            for request_future in as_completed(request_futures):
+                responses.extend((request_future.result())['embeddings'])
+        return Embeddings(responses)
```
OrenLeung, line 123: main part of change

cohere/client.py
```diff
@@ -90,21 +94,33 @@ def generate(
...
-            'texts': texts,
-            'truncate': truncate,
-        })
-        response = self.__request(json_body, cohere.EMBED_URL, model)
-        return Embeddings(response['embeddings'])
+        responses = []
+        with ThreadPoolExecutor(max_workers=self.num_workers) as executor:
+            request_futures = []
+            for i in range(0, len(texts), self.batch_size):
+                text = texts[i:i+self.batch_size]
```
jimwu6, line 113: this is probably more accurate to be `batch` or `texts_batch` or `batch_texts`?

tests/tests.py
```diff
@@ -79,6 +79,58 @@ def test_success(self):
...
+    def test_success_longer_multiple_batches(self):
+        prediction = co.embed(
+            model='small',
+            texts=['co:here', 'cohere', "embed", "python", "golang"] * 200)
+        self.assertEqual(len(prediction.embeddings), 200*5)
+        for embed in prediction.embeddings:
+            self.assertIsInstance(embed, list)
+            self.assertEqual(len(embed), 1024)
+
+    def test_success_multiple_batches_in_order(self):
```
OrenLeung, line 110: test case for that it is in order

jimwu6 (approved):
LGTM

kipply (approved):

PR #60, opened 2022-03-19
Add Extract V0 API

rodrigue-h:
This PR adds Extract to the Python SDK.

Usage:

```python
import cohere
from cohere.extract import Entity, Example

co = cohere.Client('API_KEY')

result = co.extract(
    model="large",
    examples=[
        Example(
            text="hello my name is John, and I like to play ping pong",
            entities=[
                Entity(type="Name", value="John"),
                Entity(type="Game", value="ping pong")
            ]
        ),
        Example(
            text="Jon Ke from work is amazing at chess!",
            entities=[
                Entity(type="Name", value="Jon Ke"),
                Entity(type="Game", value="chess")
            ]
        ),
        Example(
            text="David won't stop talking about playing tennis!",
            entities=[
                Entity(type="Name", value="David"),
                Entity(type="Game", value="tennis")
            ]
        ),
    ],
    texts=[
        "David will be playing soccer?"
    ]
)

print(result)
'''
cohere.Extractions {
        extractions: [
          {'id': 'caf0d948-57e8-4d48-849c-007a0c13b4cb',
          'text': 'David will be playing soccer?',
          'entities': [{'type': 'Name', 'value': 'David'}, {'type': 'Game', 'value': 'soccer'}]}]
}
'''
```


cohere/extract.py
```diff
@@ -0,0 +1,42 @@
...
+        self.iterator = iter(extractions)
+
+    def __iter__(self) -> iter:
+        return self.iterator
+
+    def __next__(self) -> next:
+        return next(self.iterator)
+
+    def __len__(self) -> int:
+        return len(self.extractions)
```
jimwu6, line 42: newline

cohere/extract.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, text: str, entities: List[ExtractEntity]) -> None:
+        self.text = text
+        self.entities = entities
+
+class Extract:
+    def __init__(self, texts: List[str], examples: List[ExtractExample]) -> None:
+        self.texts = texts
+        self.examples = examples
+    
+    def toJSON(self):
```
jimwu6, line 27: -> None

cohere/client.py
```diff
@@ -192,6 +193,26 @@ def classify(
 
         return Classifications(classifications)
 
+
+    def extract(
+        self,
+        model: str,
+        extract: Extract,
```
choicallum, line 200: is this how we want to do this in the future? this is very different from how we do the other API calls

choicallum (changes requested):

cohere/client.py
OrenLeung: yes, if we do it this way for extract, we probably wanna refactor for all

OrenLeung: i think i prefer 2 params instead of just 1 extract param

OrenLeung: `List[ExtractExample]` & `List[ExtractEntity]`

tests/tests.py
```diff
@@ -216,6 +217,42 @@ def test_success_all_fields(self):
...
+
+        extractions = co.extract(
+            'small', extract)
+
+        self.assertIsInstance(extractions, list)
+        self.assertIsInstance(extractions[0].text, str)
+        self.assertIsInstance(extractions[0].entities, list)
+        self.assertEqual(extractions[0].entities[0].type, "Name")
+        self.assertEqual(extractions[0].entities[0].value, "Roberta")
+
```
OrenLeung, line 236: add more tests
1. multiple example, each with different amount of extractentity
2. multiple example with multiple texts
3. multiple examples with multiple text but one of the text doesnt contain any extractentities

elaineg:
Can we get this reviewed? Hoping to get documentation written for this.

OrenLeung:
> Can we get this reviewed? Hoping to get documentation written for this as soon as merged.

it has been reviewed. Waiting for the suggestions to be addressed

@rodrigue-h

choicallum (approved):

tests/test_extract.py
```diff
@@ -0,0 +1,118 @@
...
+        extractions = co.extract('medium', examples, texts)
+
+        self.assertEqual(len(extractions), 2)
+        self.assertIsInstance(extractions, Extractions)
+        self.assertIsInstance(extractions[0].text, str)
+        self.assertIsInstance(extractions[1].text, str)
+        self.assertIsInstance(extractions[0].entities, list)
+        self.assertIsInstance(extractions[1].entities, list)
+        self.assertEqual(len(extractions[0].entities), 2)
+        self.assertEqual(len(extractions[1].entities), 2)
```
elliottsj, line 94: Curious: should we (can we) make assertions on the extracted string values? e.g.

```python
self.assertEqual(extractions[0].text, "Charlie")
```

Do these tests run using the live Cohere API, or some mocked data?

mkozakov: These tests are run against the Prod API, so they are more like end-to-end tests. I think in this case we can definitely do what you suggested without too much flakiness! I'll make the change

setup.py
```diff
@@ -23,7 +23,7 @@ def has_ext_modules(foo) -> bool:
 
 setuptools.setup(
     name='cohere',
-    version='1.3.8',
+    version='1.3.9',
```
elliottsj, line 26: Do we follow semantic versioning here, i.e. should this change (adding the extract API) be considered a minor update (new feature) rather than a patch?

I wonder if we should consider a more structured approach to SDK version updates. e.g. https://python-semantic-release.readthedocs.io/en/latest/ compute the version automatically based on commit messages, and generate the changelog appropriately too. Something for the future.

mkozakov: I proposed semantic versioning too! @adrian-cohere is looking at a bunch of improvements we can make to our SDKs, we're going to start a list soon.

For this case, this is a V0 API release, mostly for us to play with internally, so I think a patch is maybe applicable here?

mkozakov: I think it's a good idea. In practice, the risk here is low, unless some external party starts using an undocumented function, but there is no reason to add a non-prefixed version right now.

tests/test_extract.py
adrian-cohere: This LGTM.
Not sure if we would like to add `docblocks` on the methods. 
For me, the methods are self-explanatory, but not sure for others.

PR #72, opened 2022-05-04
Add default for 'examples' param in Client.classify method.

sabetAI:
Intellisense currently raises an error when using the `co.classify` client with finetuned models, where examples aren't (generally) necessary. 

image

Adding default param for `examples` removes the error message.


OrenLeung (approved):
LGTM

jimwu6 (approved):

mkozakov (approved):

PR #74, opened 2022-05-05
Make co.generate call async

aidangomez:
- Adds `co.batch_{generate/tokenize/detokenize}`
- Makes the Generations/Detokenize/Tokenize objects all (optionally) lazy in the sense that they can be constructed and manipulated without the REST call response they represent being complete yet. As soon as a class attribute that depends on the REST call is accessed it will block until that attribute is resolved.
- Generation object is now a subclass of `str` so that it can be thrown into code as though it were one.
- Improved the Generations object so that is can be more easily sliced/indexed into without needing to access an ugly next of attributes.
- Converted a bunch of bloat objects that were actually just NamedTuples
- Remove the "one time iterator" design decision that was made:
Previously, if you wrote:
```python
generations = co.generate(...)
for g in generations:
    print(g)
# the lines above will print the generations successfully
for g in generations:
    print(g)
# the loop above won't be executed because the iterator in generations can only be run once
```
This PR removes that strange behaviour.


cohere/client.py
```diff
@@ -33,12 +34,13 @@ class Client:
     def __init__(self,
                  api_key: str,
                  version: str = None,
-                 num_workers: int = 8,
+                 num_workers: int = 64,
```
daulet, line 37: maybe make this more universal with `multiprocessing.cpu_count()`

CoderHam, line 37: We need to be careful with making such changes. With our rate limit of 100 req/sec for trial keys a num_workers of 64 might be too much.

tests/test_generate.py
```diff
@@ -1,69 +1,56 @@
 import unittest
+
```
CoderHam, line 2: nit: remove unnecessary empty lines

cohere/generation.py
```diff
@@ -1,33 +1,70 @@
...
+
+    def _init_from_future(self, future: Future):
+        self.generations = AsyncAttribute(future, self._generations)
+
+    def _generations(self, response: Dict[str, Any]) -> List[Generation]:
+        generations: List[Generation] = []
+        for gen in response['generations']:
+            likelihood = None
+            token_likelihoods = None
+            if self.return_likelihoods == 'GENERATION' or self.return_likelihoods == 'ALL':
```
daulet, line 52: is this more pythonic? 
```suggestion
            if if self.return_likelihoods in ['GENERATION', 'ALL']:
```

tests/test_generate.py
```diff
@@ -1,69 +1,56 @@
...
-            prompt='co:here',
-            max_tokens=1)
+        prediction = co.generate(model='small', prompt='co:here', max_tokens=1)
         self.assertIsInstance(prediction.generations[0].text, str)
         self.assertIsNone(prediction.generations[0].token_likelihoods)
         self.assertEqual(prediction.return_likelihoods, 'NONE')
 
+    def test_success_batched(self):
+        _batch_size = 10
+        predictions = co.batch_generate(model='small', prompts=['co:here'] * _batch_size, max_tokens=1)
```
CoderHam, line 21: More for my understanding but in a batch_generate workflow, what if one of the prompts is too long? How would that error be returned to the user?

Additionally, how would batch_generate be rate limited? (Since it is a single request from the user)
Do we have a limit on our end on the max number of prompts that a user can specify for a specific prompt?

My concern is that a trial key user can send a batch generate call with 1M generations and it would be treated as 1 request and heavily skew our latencies for all other users

CoderHam (changes requested):

tests/test_generate.py
aidangomez: This is auto-formatted by yapf according to the specs in this repo

tests/test_generate.py
aidangomez: batch_generate is a bunch of individual generates under the hood. So rate limiting is the same as doing a bunch of indiv. calls.

For errors, if any of the parallel calls error out the program will error as expected. In a future PR I'm going to improve our error handling in python so that users can have better error handling and the errors are more informative.

cohere/client.py
aidangomez: These are threads not processes so they should be very cheap. (ie. doesn't really matter how many cores your CPU has)

As far as rate limiting, I think it should be the user's responsibility to handle retries or to wait between batches of calls. Maybe later we can add some convenience functions to make it easier, but in reality if users are hitting that barrier they should just grab a Production Key

acyrl: In `eve` we use [this](https://pypi.org/project/pyrate-limiter/) library to respect the APIs rate limit. 

If we decide to go for for the route of adding a convenience function, it is simple to use.

mkozakov:
This will need a version bump, since the error handling behavior changed

cohere/client.py
CoderHam: > but in reality if users are hitting that barrier they should just grab a Production Key

Agreed. Maybe we should encourage them to do the same. Currently it is not very obvious. An error message from rate limiting that lets them know this rate limit is for the trial key and they should use a prod key might be useful.

Please feel free to resolve

tests/test_generate.py
CoderHam: Awesome. Thank you!

CoderHam (approved):

PR #79, opened 2022-05-16
Split unit tests into separate files

mkozakov:
This PR removes the single `tests.py` file and instead divides the unit tests into individual files, one per API, to avoid an single ever-growing file


elliottsj (approved):
looks great 🍳

PR #80, opened 2022-05-16
Use correct method for string representations

mkozakov:
This PR fixes the following problem of printing out cohere objects:

**Before:**

```python
print(co.classify(...))
'''
 .cohere.Classifications {
        classifications: [<cohere.classify.Classification object at 0x1056a5bb0>]
}
'''
```
**After:**
```python
print(co.classify(...))
'''
.cohere.Classifications {
        classifications: [cohere.Classification {
        input: purple
        prediction: color
        confidence: [cohere.Confidence {
        label: fruit
        confidence: 0
}, cohere.Confidence {
        label: color
        confidence: 1
}]
}]
}
'''
```

### DIFF
--- a/cohere/response.py
+++ b/cohere/response.py
@@ -1,5 +1,5 @@
 class CohereObject():
-    def __str__(self) -> str:
+    def __repr__(self) -> str:
         contents = ''
         exclude_list = ['iterator']
 
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ class BinaryDistribution(Distribution):
 
 setuptools.setup(
     name='cohere',
-    version='1.3.7',
+    version='1.3.8',
     author='kipply',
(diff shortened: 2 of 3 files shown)


cohere/response.py
```diff
@@ -1,5 +1,5 @@
 class CohereObject():
-    def __str__(self) -> str:
+    def __repr__(self) -> str:
```
co-sudip, line 2: Should we add a test for this?

co-sudip (changes requested):

cohere/response.py
mkozakov: Good call. I added a test for classify. Adding one for embed and generate is trickier because they are not deterministic

co-sudip: I think just the one is fine.

co-sudip (approved):

PR #82, opened 2022-06-17
Add check_api_key arg to skip sanity checking api key on client init

1vn:


jimwu6 (changes requested):
update version

cohere/client.py
```diff
@@ -44,15 +44,16 @@ def __init__(self, api_key: str, version: str = None, num_workers: int = 8, requ
...
-        try:
-            res = self.check_api_key()
-            if not res['valid']:
-                raise CohereError('invalid api key')
-        except CohereError as e:
-            raise CohereError(
-                message=e.message,
-                http_status=e.http_status,
-                headers=e.headers)
+        if check_api_key:
```
co-sudip, line 47: This fix feels like a bit of a hack to be honest.  Why do we check for API key at all in init? I am assuming a request with an invalid API key will fail anyways.

co-sudip: Discussed offline. Early failure at client init makes sense in certain cases and not in others. Since we don't have a better solution, the current change looks fine.

jimwu6 (approved):

PR #84, opened 2022-07-19
Change model param to use body instead of url

lfayoux:
As a part of https://github.com/cohere-ai/blobheart/issues/3963 all endpoints will now take model as an optional param in the body instead of the URL. The old URL model will still be supported but will eventually be deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Cohere Endpoint | Function
 /classify | co.classify()
 
 ## Models
-To view an up-to-date list of available models please consult the models section in the [platform](https://os.cohere.ai). To get started try out `large` for Generate, `medium` for Classify or `small` for Embed.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ Cohere Endpoint | Function
 /classify | co.classify()
 
 ## Models
-To view an up-to-date list of available models please consult the models section in the [platform](https://os.cohere.ai). To get started try out `large` for Generate, `medium` for Classify or `small` for Embed.
+To view an up-to-date list of available models please consult the models section in the [platform](https://os.cohere.ai). To get started omit `model` from requests to use the default model sizes.
```
mkozakov, line 69: "To get started, omit" sounds a bit awkward. Maybe something like this? "When you call Cohere's APIs we decide on a good default model for your use-case behind the scenes. The default model is great to get you started, but in production environments we recommend that you specify the model size yourself via the `model` parameter. Learn more about the available models here..."

mkozakov (approved):

PR #85, opened 2022-08-03
Update SDK to use updated casing for parameters

mkozakov:
the Classify parameters `taskDescription` and `outputIndicator` are being replaced with `task_description` and `output_indicator` respectively to be consistent with the casing of other parameters.

This PR updates the SDK to use these new params.


mkozakov:
Closing this since both of these parameters will be deprecated soon anyway as a result of switching towards Embeddings based Classifications.

PR #86, opened 2022-08-09
Moderate API

1vn:


cohere/__init__.py
```diff
@@ -8,6 +8,7 @@
 EMBED_URL = 'embed'
 CLASSIFY_URL = 'classify'
 EXTRACT_URL = 'extract'
+MODERATE_URL="moderate"
```
amorisot, line 11: styling
```suggestion
MODERATE_URL= 'moderate'
```

amorisot (commented):
quickly glanced through and broadly lgtm but will let a squad member make the call 😌

cohere/client.py
```diff
@@ -199,6 +200,25 @@ def classify(
...
+        json_body = json.dumps({
+            'model': model,
+            'inputs': inputs,
+        })
+        response = self.__request(json_body, cohere.MODERATE_URL)
+
+        moderations = []
+        for res in response['results']:
+            moderations.append(
+                Moderation(res['benign'], res['profanity'], res['hate_speech'], res['violence'], res['self_harm'],
```
amorisot, line 217: are we sticking with `profanity` rather than `profane`?

1vn: we can stick with this for now since it's live

lfayoux (commented):
Do we need to bump the version number in `setup.py`?

tests/test_moderate.py
```diff
@@ -0,0 +1,18 @@
...
+from utils import get_api_key
+
+co = cohere.Client(get_api_key())
+
+
+class TestModerate(unittest.TestCase):
+
+    def test_success(self):
+        moderations = co.moderate(inputs=['I Love Cohere!'])
+        self.assertEqual(len(moderations), 1)
```
mkozakov, line 14: Can we test the contents of `moderations` in a non-flaky way? Like doing an example that obviously should fail moderation, or obviously should pass it

lfayoux (approved):

cohere/moderate.py
```diff
@@ -0,0 +1,34 @@
+from typing import List
+
+from cohere.response import CohereObject
+
+
+class Moderation(CohereObject):
```
mkozakov, line 6: This reads a bit awkward to me. Should we name this `ModerationResult`?

cohere/moderate.py
```diff
@@ -0,0 +1,34 @@
...
+        self.hate_speech = hate_speech
+        self.violence = violence
+        self.self_harm = self_harm
+        self.sexual = sexual
+        self.sexual_non_consensual = sexual_non_consensual
+        self.spam = spam
+        self.information_hazard = information_hazard
+
+
+class Moderations(CohereObject):
```
mkozakov, line 21: Do we need this wrapper? Could we just return a list of `ModerationResult`?

mkozakov: I think in the other APIs we use this wrapper so that we can attach an ID to it. So i guess either attach the response ID, or we can get rid of the wrapper.

PR #87, opened 2022-08-18
docs: Generation typehint updated to have a list of token likelihoods

eddotman:
This is a bit of a nitpick and/or trivial PR: The typehint for `Generation` has been updated from a `Likelihood` to `List[Likelihood]`. I don't think the typehints would be encountered for most users; I came across it when I wanted to mock some `co.generate()` responses in Python for unit test fixtures.


amorisot (approved):
lgtm

PR #88, opened 2022-08-19
Remove local tokenize implementation

mkozakov:
To be consistent with other SDKs, this PR removes local tokenization in favor of calling the Cohere /tokenize API.


.github/workflows/python-release.yaml
```diff
@@ -26,11 +24,6 @@ jobs:
       - name: Install dependencies
         run: yum install -y openssl-devel
 
-      - name: ✍️ auth
-        uses: google-github-actions/auth@v0
-        with:
-          credentials_json: ${{ secrets.TOKENIZER_SERVICE_ACCOUNT_KEY }}
-
```
elliottsj, line 33: Can we also remove the `Build and audit wheels` step below, and the `build-wheels.sh` script? Seems that `build-wheels.sh` compiles and uploads the tokenizer binary, but unsure if this is needed.

mkozakov: Ah i removed build-wheels.sh  but didn't catch the CI step that used it. Removed it!

elliottsj (approved):
🍳

PR #92, opened 2022-08-26
Add truncate parameter to all endpoints

1vn:


cohere/client.py
```diff
@@ -76,31 +72,24 @@ def check_api_key(self) -> Response:
...
+                 num_generations: int = 1,
+                 max_tokens: int = 20,
+                 temperature: float = 1.0,
+                 k: int = 0,
+                 p: float = 0.75,
+                 frequency_penalty: float = 0.0,
+                 presence_penalty: float = 0.0,
+                 stop_sequences: List[str] = None,
+                 return_likelihoods: str = 'NONE',
+                 truncate: str = 'NONE') -> Generations:
```
mkozakov, line 92: Since it's an optional param let's not send a value by default, just like we do with `model`
```suggestion
                 truncate: str = none) -> Generations:
```

mkozakov (approved):
Lots of formatting changes, we should enforce some kind of standard formatting

setup.py
```diff
@@ -24,7 +24,7 @@ def has_ext_modules(foo) -> bool:
 
 
 setuptools.setup(name='cohere',
-                 version='2.2.1',
+                 version='2.2.2',
```
lfayoux, line 27: Can you change to 2.2.3 I just did a release to 2.2.2

PR #101, opened 2022-09-20
add support for summarize endpoint

daulet:
Add support for summarize endpoint. Tested against local deployment.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -232,12 +232,14 @@ class Client:
 
         return Extractions(extractions)
 
-    def summarize(self, text: str) -> Summary:
+    def summarize(self, text: str, length: str, useBaseline: bool) -> Summary:
         json_body = json.dumps({
             'text': text,
+            'length': length,
+            'useBaseline': useBaseline,
         })
         response = self.__request(json_body, cohere.SUMMARIZE_URL)
-        return Summary(response['summary'])
+        return Summary(response['summary'], response['ratio'])
 
     def tokenize(self, text: str) -> Tokens:
         json_body = json.dumps({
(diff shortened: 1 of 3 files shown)


choicallum:
lgtm right now but gonna need to add other fields in response / request

cohere/summarize.py
```diff
@@ -0,0 +1,6 @@
+from cohere.response import CohereObject
+
+
+class Summary(CohereObject):
+    def __init__(self, text: str) -> None:
+        self.text = text
```
choicallum, line 6: initially looked a bit odd to me (shouldn't this field be summary?) but summary.summary also doesn't make sense and this is consistent with generate so fine by me

cohere/client.py
```diff
@@ -232,11 +232,28 @@ def unstable_extract(self, examples: List[ExtractExample], texts: List[str]) ->
...
+        self, 
+        text: str, 
+        length: str = "normal", 
+        useBaseline: bool = False,
+        max_tokens: int = 120,
+        temperature: float = 1.0,
+        k: int = 1,
+        p: float = 1,
+        frequency_penalty: float = 0.0,
+        presence_penalty: float = 0.0
```
daulet, line 245: @choicallum don't duplicate defaults here. If we change defaults in the service they'd be ignored by SDK users

aidangomez:
@daulet @choicallum do you want to update this for merger or close for now?

choicallum:
> @daulet @choicallum do you want to update this for merger or close for now?

Summarize is a cohere-only endpoint for now, part of why we wanted to push towards having an internal repo for development. We can probably close for now, but definitely no merge yet.

PR #104, opened 2022-09-30
Remove moderate from SDK

mkozakov:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.0
+* [#104](https://github.com/cohere-ai/cohere-python/pull/96)
+    * Remove experimental `moderate` api
```
eddotman, line 5: Would this be a major semver bump, since it is breaking with past versions that use `moderate`?

eddotman (changes requested):
I think we might also want to remove this [line](https://github.com/cohere-ai/cohere-python/blob/main/cohere/__init__.py#L11). 

I left a versioning question as well, but the other parts LGTM 🚀

CHANGELOG.md
eddotman: Oh unless we don't consider it breaking because the endpoint was experimental 🤔 -- pls disregard comment if so.

mkozakov: We haven't formalized it, but unofficially we talk about experimental APIs as pre-beta. That means that an experimental api can be pulled at any point without a version change

mkozakov:
> I think we might also want to remove this [line](https://github.com/cohere-ai/cohere-python/blob/main/cohere/__init__.py#L11).
> 
> I left a versioning question as well, but the other parts LGTM 🚀

Good catch! VSCode didn't save the file when i removed it :(

1vn (commented):
LGTM!

PR #107, opened 2022-10-11
remove moderate tests

choicallum:
Since co.moderate has been disabled, we should also disable the tests associated with that endpoint.


uday-cohere (approved):
thx for looking into this!

lfayoux (approved):

PR #110, opened 2022-10-19
Classification.confidence as a float not a list

mkozakov:
Before: 
```
res = co.classify(...)
res.classifications[0].confidence # [ Confidence(label: __, confidence: 0.8), Confidence(label: __, confidence: ___), ...)
```

After
```
res = co.classify(...)
res.classifications[0].confidence # 0.8
```


lfayoux:
How do you feel about changing the return type of classify to Classifications to List[Classification]. So instead of 
```
res = co.classify(...)
res.classifications[0].confidence # 0.8
```
It would be 
```
res = co.classify(...)
res[0].confidence # 0.8
```

mkozakov:
I feel great about it - Aidan already has this change in PR

lfayoux (approved):

PR #114, opened 2022-10-28
Support for transcribe endpoint

nliu96:


cohere/client.py
```diff
@@ -220,6 +221,10 @@ def detokenize(self, tokens: List[int]) -> Detokenization:
         })
         return Detokenization(_future=self._executor.submit(self.__request, json_body, cohere.DETOKENIZE_URL))
 
+    def whisper(self, file_paths: List[str]) -> Whispers:
+        response = self.__file_request(file_paths, cohere.WHISPER_URL)
```
CoderHam, line 225: Handle error when file_paths is not a list of strings

cohere/__init__.py
```diff
@@ -8,6 +8,7 @@
 EMBED_URL = 'embed'
 CLASSIFY_URL = 'classify'
 EXTRACT_URL = 'extract'
+WHISPER_URL = 'whisper'
```
CoderHam, line 11: In blobheart the endpoint is currently 'speech'
We should align to one of these

CoderHam (changes requested):

cohere/client.py
nliu96: What is the best way to check typing. You can't check if it's a list of strings with isInstance easily.

cohere/__init__.py
nliu96: Renamed to speech.

cohere/client.py
CoderHam: Yes
```
if isinstance(value, List):
```

cohere/__init__.py
```diff
@@ -8,6 +8,7 @@
 EMBED_URL = 'embed'
 CLASSIFY_URL = 'classify'
 EXTRACT_URL = 'extract'
+SPEECH_URL = 'whisper/speech'
```
CoderHam, line 11: Should this not be just "speech'?

CoderHam (changes requested):

cohere/__init__.py
CoderHam: The model name should move to the co.speech command to be symmetric with the rest of the endpoints.

nickfrosst: i think co.transcribe would be more symmetric with the other endpoints, as all our other endpoints are verbs.

CoderHam: Made that fix 🎆

cohere/__init__.py
```diff
@@ -8,6 +8,7 @@
 EMBED_URL = 'embed'
 CLASSIFY_URL = 'classify'
 EXTRACT_URL = 'extract'
+TRANSCRIBE_URL = 'whisper/transcribe'
```
jpekmez, line 11: you can use 'transcribe'
```suggestion
TRANSCRIBE_URL = 'transcribe'
```

CoderHam: Done :)

CoderHam (approved):

nliu96:
Closing since not needed right now

PR #119, opened 2022-11-04
Support for Token Streaming

abdurj:
Add token streaming support for python sdk, add `stream` and `chunk_size` optional params to `co.generate()`

**Sample use case**
```
import cohere
import os

co = cohere.Client(os.environ["CO_API_KEY"])

response_generator = co.generate("Today I went to the park", stream=True)

for gen in response_generator.stream():
   # do something with gen

final_response = response_generator.response()
```

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -102,7 +102,8 @@ class Client:
                  return_likelihoods: str = 'NONE',
                  truncate: str = None,
                  logit_bias: Dict[int, float] = {},
-                 stream: bool = False) -> Union[Generations, GenerationStream]:
+                 stream: bool = False,
+                 chunk_size: int = 1) -> Union[Generations, GenerationStream]:
         json_body = {
             'model': model,
             'prompt': prompt,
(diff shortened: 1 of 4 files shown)


cohere/generation.py
```diff
@@ -1,3 +1,5 @@
+import sseclient
```
sanderland, line 1: shouldn't this be in setup/requirements? or only imported when needed and documented as optional dependency

cohere/client.py
```diff
@@ -235,11 +242,19 @@ def __request(self, endpoint, json=None) -> Any:
...
         }
+
         if self.cohere_version != '':
             headers['Cohere-Version'] = self.cohere_version
 
         url = urljoin(self.api_url, endpoint)
-        if use_xhr_client:
+        if 'stream' in json.keys() and json['stream']:
+            headers.update({
+                'Content-Type': 'text/event-stream',
```
sanderland, line 252: this seems incorrect: the content-type of the request is still just json

sanderland:
closing as out of date

PR #128, opened 2022-12-20
Update detectlang.py

mkozakov:
Two changes in this PR:
- Fix type hinting to support older versions of python
- Use a lower version of python in CI to catch these issues


1vn (approved):
LGTM

PR #129, opened 2022-12-21
Add support for end_sequences param in Generate

mkozakov:
Adds support for the `end_sequences` param for the Generate API. It acts just like `stop_sequences`, but the sequence gets excluded from the response.


cohere/client.py
```diff
@@ -96,6 +96,7 @@ def generate(self,
                  p: float = None,
                  frequency_penalty: float = None,
                  presence_penalty: float = None,
+                 end_sequences: List[str] = None,
```
abdullahkady, line 99: why not include it in a test case? or do we not test every param?

abdullahkady (approved):

cohere/client.py
mkozakov: The way the tests are set up currently, it's hard to test parameters that alter the generation.

PR #133, opened 2023-01-11
client_name parameter to allow applications to identify themselves

sanderland:
As requested by @egrefen

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -35,13 +35,28 @@ class Client:
                  version: str = None,
                  num_workers: int = 64,
                  request_dict: dict = {},
-                 check_api_key: bool = True) -> None:
+                 check_api_key: bool = True,
+                 client_name: str = None) -> None:
+        """
+        Initialize the client.
+        Args:
+           * api_key (str): Your API key.
+           * version (str): API version to use. Will use cohere.COHERE_VERSION by default.
+           * num_workers (int): Maximal number of threads for parallelized calls.
+           * request_dict (dict): Additional parameters for calls to requests.post
+           * check_api_key (bool): Whether to check the api key for validity on initialization.
+           * client_name (str): A string to identify your application for analytics purposes.
+        """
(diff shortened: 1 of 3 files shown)


cohere/client.py
```diff
@@ -35,13 +35,28 @@ def __init__(self,
...
+                 client_name: str = None) -> None:
+        """
+        Initialize the client.
+        Args:
+           * api_key (str): Your API key.
+           * version (str): API version to use. Will use cohere.COHERE_VERSION by default.
+           * num_workers (int): Maximal number of threads for parallelized calls.
+           * request_dict (dict): Additional parameters for calls to requests.post
+           * check_api_key (bool): Whether to check the api key for validity on initialization.
+           * client_name (str): A string to identify your application for analytics purposes.
```
egrefen, line 48: Should we indicate that it's for analytics purposes _on the cohere end?_ We don't expect to give external access to this data, do we?

Developers who need to do tracking for analytics can do so in the application which encapsulates API calls made via the SDK.

sanderland: I think it does not make a lot of sense to allow it per call. If you really need to (multiple sub-applications?), you could simply instantiate multiple clients.

egrefen: Fair enough.

egrefen: Wait, this is responding to my other comment, no? What about the point about the docstring?

sanderland: Don't really have a strong opinion on that one, but added 'internal'.

egrefen (approved):
LGTM, new (and old) unit tests pass and are sensible. I guess for sanity check it would be good to check we can actually find the calls from the unit tests on our system, but I'm not sure how to do that myself.

egrefen:
I'll let @mkozakov have a look rather than merge myself.

mkozakov (approved):
LGTM, do we also want to add this to the Node + GO sdks?

PR #134, opened 2023-01-12
Add documentation for co.feedback

aidangomez:


cohere/response.py
```diff
@@ -66,5 +66,21 @@ def __repr__(self) -> str:
         output = f'cohere.{type(self).__name__} {{\n{contents}}}'
         return output
 
-    def feedback(self, feedback: str, accepted: bool) -> Feedback:
-        return self.client.feedback(id=self.id, feedback=feedback, accepted=accepted)
+    def feedback(self, good_generation: bool, feedback: str = "") -> Feedback:
```
mkozakov, line 69: This method is generation specific, can we put in Generation class, not CohereObject?

cohere/client.py
```diff
@@ -224,14 +224,32 @@ def detect_language(self, texts: List[str]) -> List[Language]:
             results.append(Language(result["language_code"], result["language_name"]))
         return DetectLanguageResponse(results)
 
-    def feedback(self, id: str, feedback: str, accepted: bool):
+    def feedback(self, id: str, good_generation: bool, feedback: str = "") -> Feedback:
```
mkozakov, line 227: I don't think we need this function in the client. It's specific to Generate, so just having it in the Generation class is a better place

aidangomez: I can definitely move this under generate, but I thought that might 1) hurt discoverability, and 2) we might want to support feedback for other endpoints in the future. Thoughts?

cohere/response.py
```diff
@@ -66,5 +66,35 @@ def __repr__(self) -> str:
         output = f'cohere.{type(self).__name__} {{\n{contents}}}'
         return output
 
-    def feedback(self, feedback: str, accepted: bool) -> Feedback:
-        return self.client.feedback(id=self.id, feedback=feedback, accepted=accepted)
+    def feedback(self, good_response: bool, desired_response: str = "", feedback: str = "") -> Feedback:
```
mkozakov, line 69: a bit confusing that the function is named feedback, and so is the param. Maybe something like this?
```suggestion
    def feedback(self, good_response: bool, desired_response: str = "", additional_feedback: str = "") -> Feedback:
```

cohere/client.py
```diff
@@ -224,14 +224,44 @@ def detect_language(self, texts: List[str]) -> List[Language]:
...
+    def feedback(self, id: str, good_response: bool, desired_response: str = "", feedback: str = "") -> Feedback:
+        """Give feedback on a response from the Cohere API to improve the model.
+
+        Can be used programmatically like so:
+
+        Example: a user accepts a model's suggestion in an assisted writing setting
+        ```
+        generations = co.generate(f"Write me a polite email responding to the one below:\n{email}\n\nResponse:")
+        if user_accepted_suggestion:
+            generations[0].feedback(good_generation=True)
```
mkozakov, line 236: wrong param name, here and in the rest of the docstrings
```suggestion
            generations[0].feedback(good_response=True)
```

cohere/client.py
```diff
@@ -224,14 +224,44 @@ def detect_language(self, texts: List[str]) -> List[Language]:
...
 
-    def feedback(self, id: str, feedback: str, accepted: bool):
+    def feedback(self, id: str, good_response: bool, desired_response: str = "", feedback: str = "") -> Feedback:
+        """Give feedback on a response from the Cohere API to improve the model.
+
+        Can be used programmatically like so:
+
+        Example: a user accepts a model's suggestion in an assisted writing setting
+        ```
+        generations = co.generate(f"Write me a polite email responding to the one below:\n{email}\n\nResponse:")
```
mkozakov, line 234: here and in other examples
```suggestion
        generations = co.generate(f"Write me a polite email responding to the one below:\n{email}\n\nResponse:").generations
```

mkozakov (approved):
Some small change requests, otherwise good to go!

cohere/client.py
aidangomez: the .generations isn't necessary anymore. We have convenience methods that let you index into the Generations object directly instead of having to access a property

1vn (approved):
LGTM

PR #136, opened 2023-01-19
fix: tests broken due to deprecated model

lfayoux:
`small` is no longer available on generate
`medium` is no longer available on classify
preset had to be changed since it was using `small` on generate

### DIFF
--- a/tests/test_classify.py
+++ b/tests/test_classify.py
@@ -9,7 +9,7 @@ co = cohere.Client(get_api_key())
 class TestClassify(unittest.TestCase):
 
     def test_success(self):
-        prediction = co.classify(model='medium',
+        prediction = co.classify(model='small',
                                  inputs=['purple'],
                                  examples=[
                                      Example('apple', 'fruit'),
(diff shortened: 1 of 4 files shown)


mkozakov (approved):

tests/test_classify.py
```diff
@@ -9,7 +9,7 @@
 class TestClassify(unittest.TestCase):
 
     def test_success(self):
-        prediction = co.classify(model='medium',
+        prediction = co.classify(model='small',
```
CoderHam, line 12: We have rules for rerouting of requests. We should understand if and how the re-routing broke. 

Approving but these changes shouldn't technically be needed

CoderHam (approved):

tests/test_classify.py
CoderHam: PS: how come we do not need to do this for embed?

lfayoux:
> should we remove `model` from the tests altogether, so that they never go stale?

we could, but we should still test that the model param works and preset request will also go stale anyways

tests/test_classify.py
lfayoux: embed still has a small model that we're using to test

PR #137, opened 2023-01-19
Add a retry policy for failed requests

lfayoux:
Large embed requests were failing due to transient errors during batching, which would fail the entire request. The retry policy should prevent the entire batch of requests from failing due to a single error.


mkozakov (approved):

mkozakov:
Thank you Lucas!!

PR #142, opened 2023-01-23
Add changes to release new version with the model param in co.chat()

lusmoura:
- Change module's version
- Add description to changelog


eddotman:
I think this should be 3.2.9 -- looks like the main version was updated in between the last PR and this one 😅

lfayoux (approved):

PR #146, opened 2023-02-01
Add experimental co.rerank API

lfayoux:
This PR adds support for the experimental co.rerank API. Access to this API is limited to a small number of users, who will be using the Python SDK.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -1,7 +1,7 @@
 import json
 import sys
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Union
 from urllib.parse import urljoin
 
 import requests
@@ -22,6 +22,7 @@ from cohere.feedback import Feedback
 from cohere.generation import Generations
 from cohere.tokenize import Tokens
 from cohere.summarize import SummarizeResponse
+from cohere.rerank import Reranking
 
 use_xhr_client = False
 try:
@@ -308,7 +309,6 @@ class Client:
         Returns:
             Feedback: a Feedback object
         """
-
         json_body = {
             'id': id,
             'good_response': good_response,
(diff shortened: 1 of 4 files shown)


cohere/rerank.py
```diff
@@ -0,0 +1,60 @@
+from typing import List, Optional, Dict, NamedTuple, Any, Iterator
+
+from cohere.response import CohereObject
+
+RerankDocument = NamedTuple("Document", [("text", str)])
```
mkozakov, line 5: let's add docs here as well via 
```
RerankDocument.__docs__ = """..."""

File: cohere/rerank.py [hunk 2]
```diff
@@ -0,0 +1,64 @@
...
+RerankDocument.__doc__ = """
+Returned by co.rerank,
+dict which always contain text but can also contain aribitrary fields
```
mkozakov, line 8:
```suggestion
dict which always contains text but can also contain aribitrary fields
```

mkozakov (approved):
Lets make sure we have docstrings for any classes that would be exposed to the user

cohere/client.py
```diff
@@ -318,6 +318,38 @@ def feedback(self, id: str, good_response: bool, desired_response: str = "", fee
         self.__request(cohere.FEEDBACK_URL, json_body)
         return Feedback(id=id, good_response=good_response, desired_response=desired_response, feedback=feedback)
 
+    def rerank(self,
+               query: str,
+               documents: Union[List[str], List[Dict[str, Any]]],
+               top_n: int = None) -> Reranking:
+        """Returns an ordered list of documents oridered by their relevance to the provided query
```
amorisot, line 325:
```suggestion
        """Returns an ordered list of documents ordered by their relevance to the provided query
```

cohere/client.py
```diff
@@ -318,6 +318,38 @@ def feedback(self, id: str, good_response: bool, desired_response: str = "", fee
...
+
+        Args:
+            query (str): The search query
+            documents (list[str], list[dict]): The documents to rerank
+            top_n (int): (optional) The number of results to return, defaults to return all results
```
amorisot, line 330:
```suggestion
            top_n (int): (optional) The number of results to return, defaults to returning all results
```

cohere/rerank.py
```diff
@@ -0,0 +1,64 @@
+from typing import List, Optional, Dict, NamedTuple, Any, Iterator
+
+from cohere.response import CohereObject
+
+RerankDocument = NamedTuple("Document", [("text", str)])
+RerankDocument.__doc__ = """
+Returned by co.rerank,
+dict which always contains text but can also contain aribitrary fields
```
amorisot, line 8:
```suggestion
dict which always contains text but can also contain arbitrary fields
```

sebastian-hofstaetter (approved):
lgtm

PR #150, opened 2023-02-03
Add support for "abstractiveness" param in co.summarize 

mkozakov:
Introduces the `abstractiveness` param for `co.summarize`. `Abstractiveness` can be set to "low", "medium", "high", and it control how close to the original text the summary is. "Low" abstractiveness summaries will lean towards reusing sentences verbatim, while "high" abstractiveness summaries will tend to paraphrase more. 

This PR also modifies the changelog to stop indicating patch versions.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -221,7 +221,7 @@ class Client:
             abstractiveness (str) One of {"high", "medium", "low"}, defaults to "high". \
                 Controls how close to the original text the summary is. "Low" abstractiveness \
                 summaries will lean towards reusing sentences verbatim, while "high" abstractiveness \
-                summaries will tend to paraphrase more. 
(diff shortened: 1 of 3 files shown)


cohere/client.py
```diff
@@ -208,7 +208,7 @@ def classify(self,
         return Classifications(classifications)
 
     def summarize(self, text: str, model: str = None, length: str = None, format: str = None, temperature: float = None,
-                  additional_instruction: str = None) -> SummarizeResponse:
+                  additional_instruction: str = None, abstractiveness: str = None) -> SummarizeResponse:
```
lfayoux, line 211: is this getting added to the json?

mkozakov: bitmoji

lfayoux (approved):

PR #154, opened 2023-02-09
Co.chat - add return chatlog

lusmoura:
## What this PR does

- Adds `return_chatlog` as an option param to co.chat
- Returns the chatlog if `return_chatlog` is set to true
- Adds unit tests for the new functionality
- Closes CNV-288

## How it was tested
- Unit tests

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -151,6 +151,34 @@ class Client:
              persona: str = "cohere",
              model: str = None,
              return_chatlog: bool = False) -> Chat:
+        """Returns a Chat object with the query reply.
+
+        Args:
+            query (str): The query to send to the chatbot.
+            session_id (str): (Optional) The session id to continue the conversation.
+            persona (str): (Optional) The persona to use.
+            model (str): (Optional) The model to use for generating the next reply.
+            return_chatlog (bool): (Optional) Whether to return the chatlog.
+
+        Example:
+        ```
+        res = co.chat(query="Hey! How are you doing today?")
+        print(res.reply)
+        print(res.session_id)
+        ```
+
+        Example:
+        ```
+        res = co.chat(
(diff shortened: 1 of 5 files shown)


cohere/client.py
```diff
@@ -145,15 +145,21 @@ def generate(self,
         response = self._executor.submit(self.__request, cohere.GENERATE_URL, json=json_body)
         return Generations(return_likelihoods=return_likelihoods, _future=response, client=self)
 
-    def chat(self, query: str, session_id: str = "", persona: str = "cohere", model: str = None) -> Chat:
+    def chat(self,
```
mkozakov, line 148: can we please add a docstring (similar to the one in summarize?)

mkozakov (approved):

mkozakov (changes requested):
also please bump the patch version in setup.py

cohere/client.py
lusmoura: Yep! On it

lusmoura:
> also please bump the patch version in setup.py

I'll also do the same for another variable, should I update the patch version twice or can it wait for the next one? @mkozakov

mkozakov:
> > also please bump the patch version in setup.py
> 
> I'll also do the same for another variable, should I update the patch version twice or can it wait for the next one? @mkozakov

If you're updating in two different PRs, update the version separately

setup.py
```diff
@@ -24,7 +24,7 @@ def has_ext_modules(foo) -> bool:
 
 
 setuptools.setup(name='cohere',
-                 version='3.3.4',
+                 version='3.4.0',
```
mkozakov, line 27: Can you please add an entry in changelog.md?

mkozakov (approved):

mkozakov (approved):

PR #158, opened 2023-02-14
Co.chat - add remaining params

lusmoura:
## What this PR does

- Adds `return_prompt` as an option param to co.chat
- Returns the prompt if `return_prompt` is set to true
- Adds `preamble_override` as an option param to co.chat
- Adds unit tests for the new functionalities
- Closes CNV-284

## How it was tested
- Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - [#158](https://github.com/cohere-ai/cohere-python/pull/158)
   - Add support for `preamble_override` parameter for co.Chat
   - Add support for `return_prompt` parameter for co.Chat
+  - Add support for `username` parameter for co.Chat
   
 ## 3.5
 - [#157](https://github.com/cohere-ai/cohere-python/pull/157)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.6
+- [#158](https://github.com/cohere-ai/cohere-python/pull/158)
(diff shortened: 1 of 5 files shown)


lfayoux (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
 # Changelog
 
-## 3.7
+## 3.8
```
sanderland, line 3: you did not update the version

cohere/client.py
```diff
@@ -153,7 +153,10 @@ def chat(self,
              persona: str = "cohere",
              model: Optional[str] = None,
              return_chatlog: bool = False,
-             chatlog_override: Optional[List[Dict[str, str]]] = None) -> Chat:
+             return_prompt: bool = False,
```
sanderland, line 156: why is this a parameter at all?

sanderland (commented):
please update version

cohere/client.py
eddotman: Prompt construction is opaque in the chat endpoint since what you pass in to the endpoint is not equal to the full prompt. 

So this is primarily to help devs debug -- we had it requested by a few users 😁.

lusmoura: just reflecting the endpoint updates, this is mainly for debugging purposes. [here](https://linear.app/cohereai/issue/CNV-287/add-return-prompt-to-the-python-sdk#comment-c9d4b7ce)'s a brief discussion about it

sanderland (approved):

PR #159, opened 2023-02-16
Fix type errors in client.py

vshmyhlo:
* Fix typing errors in `client.py`
  * A lot of arguments were typed as required (e.g. `float`, `str`), while having `None` as default value - this results in type errors, changed those to `Optional[...]`
* Fixes `json` library claashing with `json` argument in `__request` (see comment)
* Fix api key validation types 
* A bit of formatting

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -1,7 +1,7 @@
-import json
+import json as jsonlib
 import sys
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, TypedDict, Union
 from urllib.parse import urljoin
 
 import requests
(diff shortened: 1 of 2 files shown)


cohere/client.py
```diff
@@ -455,7 +462,7 @@ def __request(self, endpoint, json=None) -> Any:
 
         url = urljoin(self.api_url, endpoint)
         if use_xhr_client:
-            response = self.__pyfetch(url, headers, json.dumps(json))
+            response = self.__pyfetch(url, headers, jsonlib.dumps(json))
```
vshmyhlo, line 465: `json` argument shadows `json` library, resulting in `json.dumps` being called on request body

mkozakov (approved):

PR #160, opened 2023-02-16
SDK 4.0: AsyncClient, Sphinx docs, refactor

sanderland:
* Adds a second client: AsyncClient, which supports the asyncio pattern.
* Adds sphinx/readthedocs based documentation
* Changes formatting to black/isort/autoflake and uses poetry.
* Various refactoring


cohere/client.py
```diff
@@ -334,23 +318,20 @@ def summarize(
             additional_command (str): (Optional) Modifier for the underlying prompt, must \
```
mkozakov, line 318: Might need to update the docstring to be "Returns a generated summary" instead of "Return ..." to be consistent with the other docs

cohere/client.py
```diff
@@ -334,23 +318,20 @@ def summarize(
...
+                >>> res = co.summarize(text="Stock market report for today...")
+                >>> print(res.summary)
+
+            Summarize a text with a specific model and prompt:
+                >>> res = co.summarize(
+                >>>     text="Stock market report for today...",
+                >>>     model="summarize-xlarge",
+                >>>     length="long",
+                >>>     format="bullets",
+                >>>     temperature=0.9,
```
mkozakov, line 332:
```suggestion
                >>>     temperature=0.3,
```
As per matthias's request

cohere/client.py
```diff
@@ -368,20 +349,43 @@ def summarize(
...
 
     def batch_tokenize(self, texts: List[str]) -> List[Tokens]:
-        return [self.tokenize(t) for t in texts]
+        """A batched version of tokenize"""
+        with futures.ThreadPoolExecutor(max_workers=self.num_workers) as executor:
+            res = executor.map(self.tokenize, texts)
+        return list(res)
 
     def tokenize(self, text: str) -> Tokens:
+        """Return a Tokens object of the provided text, see https://docs.cohere.ai/reference/tokenize for advanced usage.
```
mkozakov, line 358:
```suggestion
        """Returns a Tokens object of the provided text, see https://docs.cohere.ai/reference/tokenize for advanced usage.
```

cohere/client.py
```diff
@@ -368,20 +349,43 @@ def summarize(
...
 
     def batch_detokenize(self, list_of_tokens: List[List[int]]) -> List[Detokenization]:
-        return [self.detokenize(t) for t in list_of_tokens]
+        """A batched version of detokenize"""
+        with futures.ThreadPoolExecutor(max_workers=self.num_workers) as executor:
+            res = executor.map(self.detokenize, list_of_tokens)
+        return list(res)
 
     def detokenize(self, tokens: List[int]) -> Detokenization:
+        """Return a Detokenization object of the provided tokens, see https://docs.cohere.ai/reference/detokenize-1 for advanced usage.
```
mkozakov, line 374:
```suggestion
        """Returns a Detokenization object of the provided tokens, see https://docs.cohere.ai/reference/detokenize for advanced usage.
```
The -1 is a weird artfact

cohere/client.py
```diff
@@ -368,20 +349,43 @@ def summarize(
...
+        Args:
+            tokens (List[int]): A list of tokens to convert to strings
+        """
         json_body = {'tokens': tokens}
-        return Detokenization(_future=self._executor.submit(self.__request, cohere.DETOKENIZE_URL, json=json_body))
+        res = self.__request(cohere.DETOKENIZE_URL, json=json_body)
+        return Detokenization(text=res['text'])
 
     def detect_language(self, texts: List[str]) -> DetectLanguageResponse:
+        """Return a DetectLanguageResponse object of the provided texts, see https://docs.cohere.ai/reference/detect-language-1 for advanced usage.
```
mkozakov, line 384:
```suggestion
        """Returns a DetectLanguageResponse object of the provided texts, see https://docs.cohere.ai/reference/detect-language-1 for advanced usage.
```

cohere/client.py
```diff
@@ -368,20 +349,43 @@ def summarize(
...
+        
+        Args:
+            texts (List[str]): A list of strings to identify by language
```
mkozakov, line 387:
```suggestion
            texts (List[str]): A list of texts to identify language for
```

cohere/client.py
```diff
@@ -465,61 +469,40 @@ def __print_warning_msg(self, response: Response):
...
+                backoff_factor=0.5,
+                allowed_methods=['POST', 'GET'],
+                status_forcelist=[429, 500, 502, 503, 504],
+            )
+            session.mount('https://', HTTPAdapter(max_retries=retries))
+            session.mount('http://', HTTPAdapter(max_retries=retries))
+
+            response = session.request(method, url, headers=headers, json=json, **self.request_dict)
+            try:
+                res = response.json()
```
mkozakov, line 492: Can we do something like:

```python
try:
  response = session.request(method, url, headers=headers, json=json, **self.request_dict)
except:
  raise CohereConnectionError(...)

try:
  response.raise_for_status()
except HTTPError as e:
            code = e.response.status_code
            if code >= 500:
              raise CohereServerError(...)
            elif code >= 400
              raise CohereClientError(...)

res = response.json()
```

cohere/client.py
```diff
@@ -465,61 +469,40 @@ def __print_warning_msg(self, response: Response):
...
-                        headers=response.headers,
-                    )
-                if 'message' in res:  # has errors
-                    raise CohereError(
-                        message=res['message'],
-                        http_status=response.status_code,
-                        headers=response.headers,
-                    )
-                self.__print_warning_msg(response)
+            if 'message' in res:  # has errors
```
mkozakov, line 499: not sure we need this

cohere/responses/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from cohere.responses.generation import Generation, Generations
+from cohere.responses.tokenize import Tokens, Detokenization
+from cohere.responses.classify import Classification,Classifications,LabelPrediction
+from cohere.responses.chat import Chat
+from cohere.responses.detectlang import DetectLanguageResponse,Language
+from cohere.responses.rerank import RerankDocument,Reranking,RerankResult
+from cohere.responses.summarize import SummarizeResponse
+from cohere.responses.feedback import Feedback
+from cohere.responses.cluster import ClusterJobResult,CreateClusterJobResponse,AsyncCreateClusterJobResponse
+
```
mkozakov, line 11:
```suggestion

```

mkozakov (commented):
This is a partial review, i have not finished reviewing the whole thing, but figured sending things as I go is useful

.github/workflows/python-release.yaml
```diff
@@ -1,18 +1,10 @@
-name: Python Release
+name: Release Python Package
 
 on:
   workflow_dispatch:
   push:
     branches:
       - main
-    paths:
```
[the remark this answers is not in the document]
sanderland: -> pyproject.toml and kept

.github/workflows/test.yaml
```diff
@@ -1,12 +1,37 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        pip install poetry==1.2.2
+        poetry install
+
+    # TODO: run pre-commit hook as test
```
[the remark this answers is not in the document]
sanderland: check that the code is up to the formatting standard set by the pre-commit hook

cohere/client.py
sanderland: we probably need a better status check

cohere/client.py
```diff
@@ -4,110 +4,91 @@
...
-
+    """Cohere Client
+
+    Args:
+        api_key (str): Your API key.
+        num_workers (int): Maximal number of threads for parallelized calls.
+        request_dict (dict): Additional parameters for calls to requests.post
+        check_api_key (bool): Whether to check the api key for validity on initialization.
+        client_name (str): A string to identify your application for internal analytics purposes.
+        max_retries (int): maximal number of retries for requests.
```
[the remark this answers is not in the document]
sanderland: I think it's more usual as is, a client parameter

cohere/client.py
```diff
@@ -4,110 +4,91 @@
...
+        self.api_key = api_key or os.getenv("CO_API_KEY")
         self.api_url = cohere.COHERE_API_URL
         self.batch_size = cohere.COHERE_EMBED_BATCH_SIZE
         self._executor = ThreadPoolExecutor(num_workers)
         self.num_workers = num_workers
         self.request_dict = request_dict
         self.request_source = 'python-sdk'
         self.max_retries = max_retries
+        self.timeout = timeout
+        self.cohere_version = cohere.COHERE_VERSION
```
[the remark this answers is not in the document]
sanderland: done

cohere/client.py
```diff
@@ -4,110 +4,91 @@
...
         }
-        if self.cohere_version != '':
-            headers['Cohere-Version'] = self.cohere_version
 
-        url = urljoin(self.api_url, cohere.CHECK_API_KEY_URL)
-        if use_xhr_client:
-            response = self.__pyfetch(url, headers, None)
-        else:
-            response = requests.request('POST', url, headers=headers)
+        url = posixJoin(self.api_url, cohere.CHECK_API_KEY_URL)
```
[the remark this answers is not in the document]
sanderland: @willie3838 why the different library (and strange camelCasing) here?
Also, AsyncClient doesn't seem to have this

cohere/client.py
```diff
@@ -4,110 +4,91 @@
...
+        response = requests.request('POST', url, headers=headers)
 
         try:
             res = jsonlib.loads(response.text)
         except Exception:
-            raise CohereError(message=response.text, http_status=response.status_code, headers=response.headers)
+            raise CohereAPIError.from_response(response)
         if 'message' in res.keys():  # has errors
```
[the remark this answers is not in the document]
sanderland: I think this handles cases where response.text is not valid json

cohere/client.py
```diff
@@ -126,6 +107,21 @@ def generate(self,
                  return_likelihoods: Optional[str] = None,
                  truncate: Optional[str] = None,
                  logit_bias: Dict[int, float] = {}) -> Generations:
+        """Generate endpoint.
+        See https://docs.cohere.ai/reference/generate for advanced arguments
+
+        Args:
+            prompt (str): either a single prompt, or a list of them
+            model (str): (Optional) The model to use for generating the next reply.
+            return_likelihoods (str): (Optional) One of GENERATION|ALL|NONE to specify how and if the token likelihoods are returned with the response.
```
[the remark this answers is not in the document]
sanderland: ->  token (log) likelihoods

cohere/client.py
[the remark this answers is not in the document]
sanderland: fixed asyncclient

cohere/client.py
sanderland: Tried something like this, but the exceptions raised by the different libraries are quite byzantine

cohere/client.py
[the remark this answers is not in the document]
willie3838: urljoin only takes 2 parameters when I tried it, and when i tried chaining the urljoin it'd produce an unexpected result. i can fix the camelCasing, i was too accustomed to golang's way of writing stuff. oop, i see you already did it, thanks

willie3838:
Note that the synchronous chat tests are failing for some reason? Need to address that before merging.

lfayoux:
> Note that the synchronous chat tests are failing for some reason? Need to address that before merging.

There were some breaking changes to the chat API that weren't updated in the sdk, I had to remove some tests in the 3.X version.

readthedocs.yml
```diff
@@ -0,0 +1,7 @@
+---
+build:
+  image: latest
+
+python:
+  version: 3.8
+  pip_install: true
```
mkozakov, line 7:
```suggestion
  pip_install: true

```

cohere/utils.py
```diff
@@ -0,0 +1,33 @@
...
+            elif isinstance(obj, np.floating):
+                return float(obj)
+            elif isinstance(obj, np.ndarray):
+                return obj.tolist()
+            elif isinstance(obj, set):
+                return list(obj)
+            else:
+                return super().default(obj)
+except:
+    class NumpyAwareJsonEncoder(json.JSONEncoder):
```
mkozakov, line 23: Should we name this `CustomJsonEncoder` since it's not always numpy aware?

cohere/responses/tokenize.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @property
+    def length(self):
+        return len(self)
+
+    def __len__(self) -> int:
+        return len(self.tokens)
+
+    def visualize(self, **kwargs):
+        import pandas as pd
```
mkozakov, line 36: should we wrap this in a try/except and provide a custom error message?

cohere/responses/__init__.py
```diff
@@ -0,0 +1,10 @@
+from cohere.responses.embeddings import Embedding,Embeddings
+from cohere.responses.generation import Generation, Generations
+from cohere.responses.tokenize import Tokens, Detokenization
+from cohere.responses.classify import Classification,Classifications,LabelPrediction
+from cohere.responses.chat import Chat
+from cohere.responses.detectlang import DetectLanguageResponse,Language
+from cohere.responses.rerank import RerankDocument,Reranking,RerankResult
+from cohere.responses.summarize import SummarizeResponse
+from cohere.responses.feedback import Feedback
+from cohere.responses.cluster import ClusterJobResult,CreateClusterJobResponse,AsyncCreateClusterJobResponse
```
mkozakov, line 10:
```suggestion
from cohere.responses.cluster import ClusterJobResult,CreateClusterJobResponse,AsyncCreateClusterJobResponse

```

cohere/logging.py
```diff
@@ -0,0 +1,32 @@
...
+
+def log_warning():
+    logger.setLevel(logging.WARNING)
+
+
+def log_error():
+    logger.setLevel(logging.ERROR)
+
+
+def log_off():
```
mkozakov, line 31: should this be `log_critical`?

cohere/__init__.py
```diff
@@ -1,8 +1,12 @@
-from .client import Client
-from .error import CohereError
+from cohere.client import Client
+from cohere.client_async import AsyncClient
+from cohere.error import CohereError, CohereAPIError,CohereConnectionError
 
 COHERE_API_URL = 'https://api.cohere.ai'
-COHERE_VERSION = '2022-12-06'
+RETRY_STATUS_CODES = [429, 500, 502, 503, 504]
```
mkozakov, line 6: does the retry policy handle 429 correctly? You need to wait for one minute before retrying i believe

tests/async/test_async_client.py
```diff
@@ -0,0 +1,27 @@
...
+
+@pytest.mark.asyncio
+async def test_async_enter_invalid_key():
+    cli = cohere.AsyncClient(api_key="invalid") 
+    with pytest.raises(cohere.CohereError) as exc:
+        async with cli as co:
+            pass
+    assert "invalid API key" in str(exc)
+
+
```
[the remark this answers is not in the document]
sanderland: I have not run the formatter, since changing to black will make the diff even more of a nightmare than it is. don't worry about formattting, it will run after you're done.

cohere/responses/tokenize.py
sanderland: repr_html uses try/except (and returns a reasonable alternative), and this is only used there.

cohere/__init__.py
sanderland: It does the same thing on all those error codes. Where is the 429 behaviour documented (preferably publically)

tests/sync/test_chat.py
```diff
@@ -131,25 +112,24 @@ def test_return_prompt_false(self):
...
-        with self.assertRaises(cohere.CohereError):
-            co.chat("Yo what up?", preamble_override=123).reply
+        with self.assertRaises(cohere.CohereError) as e:
+            co.chat("Yo what up?", persona_prompt=123).reply
+        self.assertIn('invalid type', str(e.exception)) 
 
     def test_username_override(self):
         username = "CustomUser"
-        prediction = co.chat("Yo what up?", username=username, return_chatlog=True)
+        prediction = co.chat("Yo what up?", user_name=username, return_chatlog=True)
```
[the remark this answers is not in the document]
sanderland: @madelinehjenkins @eddotman ?
The chat changes are simply due to the API changing under our feet and breaking tests.
On a broader note: it's not clear to me what endpoints are 'allowed' to change within the same API/SDK version and how we indicate this to users.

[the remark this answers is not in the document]
madelinehjenkins: As a general note, we're expecting more breaking changes with the `co.chat` endpoint over the coming month or so as it is in the experimental phase, but it will be more stable once we hit the alpha phase. So this may be a temporary situation, at least for chat.

As for this specific change from `username` to `user_name`, I don't know exactly what motivates it. My guess would be `username` feels more like a specific term for login credentials, while `user_name` more generally indicates the name of a user (more accurate in this case).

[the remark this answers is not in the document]
sanderland: a wait I thought this was about the main function call which was already renamed

cohere/__init__.py
mkozakov: I think we need to handle it separately:

```
if response.status == 429:
  sleep(60)
  make_request()
```

Unfortunately we don't have public documentation for the rate limiting. I just requested it again: https://cohereai.slack.com/archives/C02L95WS1B2/p1678767941930239

The way it works is that depending on what type of API key you are using and what endpoint you are hitting you are allowed X number of requests per minute, if you exceed that amount, you start receiving 429 status code and need to wait for 1 minute before you can make more requests.

Here are the limits per endpoint / key types: https://github.com/cohere-ai/blobheart/blob/main/config/limits/limits.go

sanderland: > and need to wait for 1 minute before you can make more requests.

This is what I'm not sure about. I thought if you have a limit of 100/min, and you hit the limit at 50s by making 120 calls/min, you can continue making calls by whenever your rolling window of the last 60s doesn't have 100 calls -- which is in like 1-2s, not 60s.
I know there are headers which suggest when to make the next call, but these assume no concurrency I believe.

In any case I think the asyncio client here is at feature parity with the sync one, that is, they both don't handle this perfectly.

sanderland: added sleep 5

mkozakov (approved):

PR #161, opened 2023-02-16
Add clustering endpoints

vshmyhlo:
Add create-job and get-job methods for new clustering endpoints

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -1,4 +1,5 @@
 import json as jsonlib
+import os
 import sys
 from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, List, Optional, TypedDict, Union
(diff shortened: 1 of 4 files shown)


cohere/client.py
```diff
@@ -494,3 +496,28 @@ def __request(self, endpoint, json=None) -> Any:
                 self.__print_warning_msg(response)
 
         return res
+
+    def create_cluster_job(
+        self,
+        embeddings_url: str,
+        threshold: Optional[float] = None,
+        min_cluster_size: Optional[int] = None,
```
amorisot, line 504: why `Optional`? do we have values set in the backend if these aren't specified??

vshmyhlo: there are default in backend

lfayoux (approved):

PR #162, opened 2023-02-16
fix typo in client.py

eltociear:
cointains -> contains


mkozakov (commented):
Thanks for catching it! Can you please bump the [patch version in this file?](https://github.com/cohere-ai/cohere-python/blob/main/setup.py#L27)

mkozakov (approved):

PR #163, opened 2023-02-17
Clustering: job polling

vshmyhlo:
Adds job polling methods for clustering jobs.

```python
job = co.create_cluster_job(input_file)
result = job.poll(timeout=60, interval=5)
assert result.status == 'complete'
```

or, if you stored ID somewhere and want to poll in another process

```python
job = co.create_cluster_job(input_file)
result = co.poll_cluster_job(job.job_id, timeout=60, interval=5)
assert result.status == 'complete'
```


amorisot:
this is working for me :)))

cohere/client.py
```diff
@@ -503,21 +504,83 @@ def create_cluster_job(
         threshold: Optional[float] = None,
         min_cluster_size: Optional[int] = None,
     ) -> CreateClusterJobResponse:
+        """Create clustering job.
+
+        Args:
+            embeddings_url (str): File with embeddings to cluster.
+            threshold (Optional[float], optional): TODO. Defaults to None.
```
amorisot, line 511: why the TODO?

cohere/client.py
```diff
@@ -503,21 +504,83 @@ def create_cluster_job(
...
+        self,
+        job_id: str,
+        timeout: float = 0,
+        interval: float = 10,
+    ) -> GetClusterJobResponse:
+        """Poll clustering job results.
+
+        Args:
+            job_id (str): Clustering job id.
+            timeout (float, optional): Poll timeout in seconds. Defaults to 0.
```
amorisot, line 566: why set 0 by default here? it's not intuitive to me, might be good to justify?

cohere/client.py
vshmyhlo: not sure how to word it properly, can you please help with description?

cohere/client.py
vshmyhlo: agree, updated to be `None` when no timeout, and added description to docstring

cohere/client.py
amorisot: how's this?
```suggestion
            threshold (Optional[float], optional): Similarity threshold above which two texts are deemed to belong in the same cluster. Defaults to None.
```

vshmyhlo: very nice!

PR #164, opened 2023-02-17
Cluster: add list jobs method

vshmyhlo:


cohere/cluster.py
```diff
@@ -47,3 +80,26 @@ def wait(
         """
 
         return self._wait_fn(job_id=self.job_id, timeout=timeout, interval=interval)
+
+
+def build_cluster_job_response(response: Dict[str, Any]) -> ClusterJobResult:
```
sanderland, line 85: classmethod: ClusterJobResult.from_response?

cohere/client.py
```diff
@@ -544,11 +544,17 @@ def get_cluster_job(
...
 
         response = self.__request(os.path.join(cohere.CLUSTER_JOBS_URL, job_id), method='GET')
-        return ClusterJobResult(
-            status=response['status'],
-            output_clusters_url=response['output_clusters_url'],
-            output_outliers_url=response['output_outliers_url'],
-        )
+        return build_cluster_job_response(response)
+
+    def get_cluster_jobs(self) -> List[ClusterJobResult]:
```
sanderland, line 549: a bit too similar to `get_cluster_jobs`, maybe `list_cluster_jobs` ?

sanderland (changes requested):
left some suggestions. Also assuming this is the last of the cluster PRs, please bump the minor version number

cohere/client.py
vshmyhlo: renamed

cohere/cluster.py
vshmyhlo: updated

sanderland (approved):

mkozakov (approved):

PR #166, opened 2023-02-21
Add high-level polling to cluster

vshmyhlo:
Adds job polling methods for clustering jobs.

```python
job = co.create_cluster_job(input_file)
result = job.wait(timeout=60, interval=5)
assert result.status == 'complete'
```

or, if you stored ID somewhere and want to poll in another process

```python
job = co.create_cluster_job(input_file)
result = co.wait_for_cluster_job(job.job_id, timeout=60, interval=5)
assert result.status == 'complete'
```

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -521,7 +521,7 @@ class Client:
         response = self.__request(cohere.CLUSTER_JOBS_URL, json=json_body)
         return CreateClusterJobResponse(
             job_id=response['job_id'],
-            poll_fn=self.poll_cluster_job,
+            wait_fn=self.wait_cluster_job,
         )
 
     def get_cluster_job(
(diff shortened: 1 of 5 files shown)


cohere/client.py
```diff
@@ -499,25 +500,85 @@ def create_cluster_job(
...
 
         response = self.__request(os.path.join(cohere.CLUSTER_JOBS_URL, job_id), method='GET')
-        return GetClusterJobResponse(
+        return ClusterJobResult(
             status=response['status'],
             output_clusters_url=response['output_clusters_url'],
             output_outliers_url=response['output_outliers_url'],
         )
+
+    def poll_cluster_job(
```
sanderland, line 553: with this name (and the one below) I would think it just gets the status once. Maybe `wait_for_cluster_job` and `wait_for_result` or something?

tests/test_cluster.py
```diff
@@ -26,3 +28,56 @@ def test_cluster_job(self):
...
+            min_cluster_size=3,
+            threshold=0.5,
+        )
+
+        job = create_res.poll(timeout=60, interval=5)
+        assert job.status == 'complete'
+        assert job.output_clusters_url is not None
+        assert job.output_outliers_url is not None
+
+    def test_handler_poll_times_out(self):
```
sanderland, line 72: are you very sure this will time out 100% of the time (and the other does not time out ever) ? If not, maybe mark it as skip in CI?

cohere/client.py
vshmyhlo: makes sense, renamed to `wait`

sanderland (approved):
presume you will bump the version when all these cluster things are complete

tests/utils.py
```diff
@@ -5,3 +5,8 @@ def get_api_key() -> str:
     api_key = os.getenv('CO_API_KEY')
     assert api_key, 'CO_API_KEY environment variable not set'
     return api_key
+
+
+def in_ci() -> bool:
+    ci = os.getenv('CI')
+    return ci == 'true'
```
sanderland, line 12: I like in  .lower() ['true','1'] here

vshmyhlo: nice! updated

sanderland (approved):

sanderland (approved):

PR #168, opened 2023-02-22
allow model param for rerank

lfayoux:
We're going to be deploying multiple rerank models soon, sdk needs to allow for models to be passed in for rerank.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -403,12 +403,14 @@ class Client:
     def rerank(self,
                query: str,
                documents: Union[List[str], List[Dict[str, Any]]],
+               model: str = None,
                top_n: Optional[int] = None) -> Reranking:
         """Returns an ordered list of documents ordered by their relevance to the provided query
 
         Args:
             query (str): The search query
             documents (list[str], list[dict]): The documents to rerank
+            model (str): (Optional) The model to use for re-ranking
             top_n (int): (optional) The number of results to return, defaults to returning all results
         """
         parsed_docs = []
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -24,7 +24,7 @@ def has_ext_modules(foo) -> bool:
 
 
 setuptools.setup(name='cohere',
-                 version='3.7.0',
+                 version='3.7.1',
```
sanderland, line 27: nit: this is not a bugfix, so use 3.8

sanderland (approved):

abdullahkady (approved):

cohere/client.py
```diff
@@ -403,12 +403,14 @@ def feedback(self, id: str, good_response: bool, desired_response: str = "", fee
...
                query: str,
                documents: Union[List[str], List[Dict[str, Any]]],
+               model: str = None,
                top_n: Optional[int] = None) -> Reranking:
         """Returns an ordered list of documents ordered by their relevance to the provided query
 
         Args:
             query (str): The search query
             documents (list[str], list[dict]): The documents to rerank
+            model (str): (Optional) The model to use for re-ranking
```
co-sylvie, line 413: could we say what the defaults are for these args in the doc string?

lfayoux: the default is none

co-sylvie: i mean the default behavior, if it's none we default to the english model right?

lfayoux: yeah but the default model could change, it will be set in the docs once this goes public

co-sylvie (approved):
✨

PR #169, opened 2023-02-22
[WIP] Generation streaming

sanderland:


cohere/client.py
```diff
@@ -111,7 +111,8 @@ def generate(self,
                  stop_sequences: Optional[List[str]] = None,
                  return_likelihoods: Optional[str] = None,
                  truncate: Optional[str] = None,
-                 logit_bias: Dict[int, float] = {}) -> Generations:
+                 logit_bias: Dict[int, float] = {},
+                 stream: bool = False) -> Union[Generations,StreamingGenerations]:
```
daulet, line 115:
```suggestion
                 stream: bool = False,
                 ) -> Union[Generations,StreamingGenerations]:
```
nit: this way we don't have to update two lines when we append new argument

cohere/client_async.py
```diff
@@ -127,8 +127,13 @@ async def generate(
...
             "logit_bias": logit_bias,
         }
-        response = await self.__request(cohere.GENERATE_URL, json_body)
-        return Generations.from_dict(response=response,return_likelihoods=return_likelihoods)
+        if stream:
+            json_body['stream'] = True
+        response = await self.__request(cohere.GENERATE_URL, json=json_body, stream=stream)
+        if stream:
+            return StreamingGenerations(response) 
+        else:
```
daulet, line 135: redundant else?

cohere/responses/generation.py
```diff
@@ -114,3 +115,22 @@ def prompt(self) -> str:
...
+
+
+class StreamingGenerations(CohereObject):
+
+    def __init__(self, response):
+        self.response = response
+
+    def __iter__(self) -> Generator[StreamingTokenLikelihood, None, None]:
+        if not isinstance(self.response, requests.Response):
+            raise ValueError("For AsyncClient, use `async for` to iterate through the `StreamingGenerations`")
```
daulet, line 129: do we need symmetrical check in async version?

daulet (commented):
🙌 LGTM so far

cohere/responses/generation.py
```diff
@@ -114,3 +115,22 @@ def prompt(self) -> str:
         """Returns the prompt used as input"""
         return self[0].prompt  # should all be the same
 
+
+StreamingTokenLikelihood = NamedTuple("TokenLikelihood", [("index", int), ("token", str), ("likelihood", float)])
```
daulet, line 119: not sure about my python - are these strongly typed fields or are they accessed by name like a dict?

cohere/responses/generation.py
```diff
@@ -114,3 +115,22 @@ def prompt(self) -> str:
...
+
+
+class StreamingGenerations(CohereObject):
+
+    def __init__(self, response):
+        self.response = response
+
+    def __iter__(self) -> Generator[StreamingTokenLikelihood, None, None]:
```
daulet, line 127: I wonder if we should expose an iterator per generation, i.e. grouped by `StreamingTokenLikelihood.Index`

cohere/client.py
sanderland: formatter will take care of this when 4.0 is merged

cohere/responses/generation.py
sanderland: I think this mistake is far more likely

cohere/responses/generation.py
sanderland: not sure how 🤔

cohere/client_async.py
sanderland: more readable imo

cohere/responses/generation.py
daulet: like a separate method that returns a list of iterators

daulet:
I propose we merge this without "final message" since that won't be out for a while: https://cohereai.slack.com/archives/C031A1G0PDK/p1677716021485729

cohere/responses/generation.py
daulet, line 119: we changed "token" with "text" - both returned right now, but "token" will disappear soon

sanderland: I saw your message, will fix everything up and complete this after #160 is merged.

cohere/responses/generation.py
sanderland: There isn't really a good way to do this without consuming tokens or essentially not-streaming.

lfayoux (changes requested):
Can we also add a test for streaming generation pls

lfayoux (approved):

mkozakov:
I think we are about to release changes on the backend for token streaming, so maybe worth it to hold off until this is in: https://linear.app/cohereai/issue/PTS-4100/return-status-message-per-generation

cohere/responses/generation.py
```diff
@@ -123,3 +126,38 @@ def visualize(self, **kwargs) -> str:
...
+
+    def _make_response_item(self, line) -> StreamingText:
+        streaming_item = json.loads(line)
+        index = streaming_item.get("index", 0)
+        text = streaming_item["text"]
+        id = streaming_item.get("id")
+        while len(self.texts) <= index:
+            self.texts.append("")
+            self.ids.append(None)
+        self.texts[index] += text
```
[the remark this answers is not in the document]
daulet: on the backend all generations are treated independently, order at which they start generating is not guaranteed.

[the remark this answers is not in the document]
sanderland: `while`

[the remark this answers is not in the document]
daulet: ah, all good

PR #170, opened 2023-02-22
Removed ability to override API version

willie3838:
### What this PR does.
Users can no longer pass in a version to the SDK. We want to support this because the SDK code is only specific to the API version it ships with. By default, the version is set to v1 which is equivalent to the 2022 version we currently have.

Closes OS-44
### Testing
Current unit tests passed. Remove the unit test testing for invalid version since we're presetting that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ class BinaryDistribution(Distribution):
 
 
 setuptools.setup(name='cohere',
-                 version='3.7.0',
+                 version='5.0.0',
                  author='1vn',
                  author_email='ivan@cohere.ai',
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -24,7 +24,7 @@ def has_ext_modules(foo) -> bool:
 
 
 setuptools.setup(name='cohere',
-                 version='3.7.0',
+                 version='4.0.0',
```
jimwu6, line 27: version doesn't match up with  CHANGELOG.md (which has 3.8)

sanderland: also note #160 is going to 4.0 as well so please coordinate

PR #172, opened 2023-02-28
dont raise exception on retry limit, preserve original error msg

lfayoux:
preserve the original error, 500 error will now result in:
`cohere.error.CohereError: internal server error, this has been reported to our developers`

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -502,6 +502,7 @@ class Client:
                     backoff_factor=0.5,
                     allowed_methods=['POST', 'GET'],
                     status_forcelist=[429, 500, 502, 503, 504],
+                    raise_on_status=False,
                 )
                 session.mount('https://', HTTPAdapter(max_retries=retries))
                 session.mount('http://', HTTPAdapter(max_retries=retries))
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -512,7 +512,8 @@ class Client:
                 res = response.json()
             except requests.exceptions.ConnectionError:
                 raise CohereError(
-                    message='A Connection error occurred when trying to connect to the Cohere API. Please check your internet connection.'
+                    message='A Connection error occurred when trying to connect to the Cohere API.'
+                            ' Please check your internet connection.'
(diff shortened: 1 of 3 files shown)


cohere/client.py
```diff
@@ -503,6 +503,7 @@ def __request(self, endpoint, json=None, method='POST') -> Any:
                     backoff_factor=0.5,
                     allowed_methods=['POST', 'GET'],
                     status_forcelist=[429, 500, 502, 503, 504],
+                    raise_on_status=False,
```
mkozakov, line 506: I think what we should go is still raise on status but then catch the dofferent error types. The issue with the current implementation is that it assumes that you got a response back, but if you have a connection error it will crap out

sanderland: In the 3.9 branch this change seems fine, but it would be good if @lfayoux could go over the setup in the 4.0 version as well to see what the effects are.

lfayoux: If we raise on status and catch an error then we would lose the response message that we get back. I can a try to the request to catch connection errors though

mkozakov: i think we should do it like this:

```python
with requests.Session() as session:
    retries = Retry(
        total=3,
        backoff_factor=0.5,
        allowed_methods=['POST', 'GET'],
        status_forcelist=[429, 500, 502, 503, 504],
        raise_on_status=False,
    )
    session.mount('https://', HTTPAdapter(max_retries=retries))
    session.mount('http://', HTTPAdapter(max_retries=retries))

try:
    response = session.request(method, ...)
    response.raise_for_status()
except requests.exceptions.HTTPError as http_error:
    # check response.status_code and handle appropriately
except requests.exceptions.ConnectionError as connection_error:
    # Handle connection error
except requests.exceptions.Timeout as timeout_error:
    # Handle timeout error
except requests.exceptions.RequestException as request_error:
    # Handle any other error
```

mkozakov (approved):

sanderland:
Please rebase on #160

PR #177, opened 2023-03-08
Fix chat API, merge changes, fix type for meta

sanderland:
Note this is going into #160, not main

### DIFF
--- a/cohere/responses/cluster.py
+++ b/cohere/responses/cluster.py
@@ -82,15 +82,15 @@ class ClusterJobResult(CohereObject):
             output_outliers_url=data['output_outliers_url'],
             clusters=clusters,
             error=data.get('error'),
-            meta=data['meta']
+            meta=data.get('meta')
         )
 
 
 class CreateClusterJobResponse(CohereObject):
     job_id: str
-    meta: Meta
+    meta: Optional[Meta]
 
-    def __init__(self, job_id: str, meta: Meta, wait_fn):
+    def __init__(self, job_id: str, meta: Optional[Meta], wait_fn):
         self.job_id = job_id
         self._wait_fn = wait_fn
         self.meta = meta
@@ -100,7 +100,7 @@ class CreateClusterJobResponse(CohereObject):
         return cls(
             job_id=data['job_id'],
             wait_fn=wait_fn,
-            meta=data['meta']            
+            meta=data.get('meta')
         )
 
     def wait(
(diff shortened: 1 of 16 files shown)


tests/sync/test_chat.py
```diff
@@ -26,39 +26,39 @@ def test_multi_replies(self):
             prediction = prediction.respond("oh that's cool")
             self.assertIsInstance(prediction.reply, str)
             self.assertIsInstance(prediction.session_id, str)
-            self.assertEqual(prediction.persona, "cohere")
+            self.assertEqual(prediction.persona_name, "cohere")
 
+    @unittest.skip("Not raising an error")
```
sanderland, line 31: @madelinehjenkins could you track this down?

cohere/responses/cluster.py
```diff
@@ -65,39 +67,40 @@ def __init__(
...
         )
 
 
 class CreateClusterJobResponse(CohereObject):
     job_id: str
-    meta: List[Meta]
+    meta: Meta
 
-    def __init__(self, job_id: str, meta: List[Meta], wait_fn):
+    def __init__(self, job_id: str, meta: Meta, wait_fn):
```
willie3838, line 93: this is a my bad, but this should be optional. all meta tags are optional since there are cases when it might not be sent

cohere/responses/summarize.py
```diff
@@ -2,7 +2,7 @@
 
 from cohere.responses.meta_response import Meta
 
-SummarizeResponse = NamedTuple("SummarizeResponse", [("id", str), ("summary", str), ("meta", List[Meta])])
+SummarizeResponse = NamedTuple("SummarizeResponse", [("id", str), ("summary", str), ("meta", Meta)])
```
willie3838, line 5: another my bad, meta should be optional here too.

tests/sync/test_chat.py
```diff
@@ -26,39 +26,35 @@ def test_multi_replies(self):
...
-        prediction = co.chat("Yo what up?", persona="wizard")
+        prediction = co.chat("Yo what up?", persona_name="Wizard",return_chatlog=True)
         self.assertIsInstance(prediction.reply, str)
         self.assertIsInstance(prediction.session_id, str)
-        self.assertEqual(prediction.persona, "wizard")
+        self.assertEqual(prediction.persona_name, "Wizard")
+        # TODO: check 'Wizard:' in chatlog
 
-    @unittest.skipIf(in_ci(), "brittle test")
+    @unittest.skip("Very brittle test")
```
madelinehjenkins, line 38: should we just delete this test?

madelinehjenkins (commented):
The changes to chat look good to me, left a few very minor comments.

Just commenting rather than approving as I'm not familiar with how meta is used here.

PR #179, opened 2023-03-13
Update chat params

lusmoura:
## What this PR does

- Adds `temperature` as an option param to co.chat
- Adds `max_tokens` as an option param to co.chat
- Removes `preamble_override`
- Fixes the `user_name` param 

## How it was tested
- Unit tests

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -154,7 +154,6 @@ class Client:
              return_chatlog: bool = False,
              return_prompt: bool = False,
              chatlog_override: List[Dict[str, str]] = None,
-             preamble_override: str = None,
              username: str = None) -> Chat:
         """Returns a Chat object with the query reply.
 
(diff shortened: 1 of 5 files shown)


cohere/client.py
```diff
@@ -154,8 +154,9 @@ def chat(self,
              return_chatlog: bool = False,
              return_prompt: bool = False,
              chatlog_override: List[Dict[str, str]] = None,
-             preamble_override: str = None,
-             username: str = None) -> Chat:
+             username: str = None,
```
sanderland, line 157: in #160 this is implemented as `user_name`

cohere/client.py
```diff
@@ -154,8 +154,9 @@ def chat(self,
              return_chatlog: bool = False,
              return_prompt: bool = False,
              chatlog_override: List[Dict[str, str]] = None,
-             preamble_override: str = None,
```
lusmoura, line 157: [ref](https://github.com/cohere-ai/blobheart/commit/3f4c58c895fc9b75fccdb8b985d543a2783d5adc)

cohere/client.py
lusmoura: should I change it here as well or is it better to just wait for #160 to be merged?

sanderland:
Handled in #160

PR #184, opened 2023-03-15
respect timeout option for sync client

lfayoux:
Include the timeout option in the sync request


harry-cohere (approved):

sanderland:
lgtm but install pre-commit as described in readme to fix formatting

PR #192, opened 2023-03-26
Fix duplicate call in client.py

dennlinger:
Currently there is a significant problem that causes especially problems for trial API users with limited rates.
Instead of a single access of the cohere endpoint, a request is issued *twice* instead, which means that rate limits are effectively cut in half.

Additionally, this also causes server load to be doubled, which I believe should be fixed ASAP for the interest of everyone involved ;-)

Removing either the suggested line (or the other call just below the affected line) should fix this issue.


mkozakov:
@dennlinger Thank you for catching and fixing this bug!

PR #194, opened 2023-03-27
Chat - Return generation ID

lusmoura:
### What this PR does
- Adds the generation ID to the response

### How was it tested
- Unit tests


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cohere"
-version = "4.1.2"
+version = "4.2.0"
```
harry-cohere, line 3: I don't think this change is considered breaking (because the `id` property already exists on the `Chat` object because it's inherited from the base `CohereObject` class) and therefore the version bump should be `4.1.3`

lusmoura: Oh, sorry, I thought Patch releases were only for bug fixes

PR #199, opened 2023-04-04
Chat - Add deprecation warnings

lusmoura:
### What this PR does
- Reflect API response/request changes (https://github.com/cohere-ai/blobheart/pull/8395)

### How it was tested
- Unit tests


cohere/client.py
```diff
@@ -214,16 +221,38 @@ def chat(
         if chatlog_override is not None:
             self._validate_chatlog_override(chatlog_override)
 
+        if session_id != "":
+            conversation_id = session_id
+            warnings.warn(
```
lfayoux, line 226: I think you can use `logger.warning` instead

lusmoura: I'll update =D but just out of curiosity, should we use one over the other?

lfayoux: so we can have more control over logging if needed.

lfayoux (approved):

lfayoux (approved):

PR #205, opened 2023-04-13
feat: add max_chunks_per_doc param to rerank

lfayoux:
Adds new max_chunks_per_doc param to rerank + enforces model param for rerank

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cohere"
-version = "4.1.5"
+version = "4.1.6"
 description = ""
 authors = ["Cohere"]
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


abdullahkady (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cohere"
-version = "4.1.4"
+version = "4.1.6"
```
abdullahkady, line 3: why? should be 4.1.5 right?

sanderland: 4.1.5 just went in though (:

abdullahkady: so fast haha 😂

PR #206, opened 2023-04-13
Cluster: Update cluster parameters 

alekhya-n:
Remove `threshold`  as a parameter from cluster and update tests with `n_neighbors` and `is_deterministic` params.


cohere/client.py
```diff
@@ -597,25 +597,29 @@ def _request(self, endpoint, json=None, method="POST", stream=False) -> Any:
...
 
         Args:
             embeddings_url (str): File with embeddings to cluster.
-            threshold (Optional[float], optional): Similarity threshold above which two texts are deemed to belong in
                 the same cluster. Defaults to None.
-            min_cluster_size (Optional[int], optional): Minimum number of elements in a cluster. Defaults to None.
-
+            n_neighbors (Optional[int], optional): Number of neighbors in dimensionality reduction with UMAP. Lower
+                values focus UMAP on local structure of the data while higher values will focus UMAP on global
+                structure of the data. Defaults to None.
```
sanderland, line 611: Do we want to tell people what None translates to?

alekhya-n:
> Can you investigate and fix the cluster tests failing?

Yes - this PR is not yet ready. It depends on some params being exposed in blobheart and that PR is in progress.

cohere/client.py
evren-cohere: why no make the default value the integer values used by default instead of using None?
for n_neighbors we can say "Number of nearest neighbors used by UMAP to establish the local structure of the data" and link to https://umap-learn.readthedocs.io/en/latest/parameters.html#n-neighbors

evren-cohere: @sanderland is there a usual method to specify defaults values in the API (blobheart code) and access them here via na import (e.g. from cohere.responses.cluster import ClusterDefaults)?  That way we can change in just 1 place?

tests/async/test_async_cluster.py
```diff
@@ -18,7 +18,8 @@ async def test_async_create_cluster_job(async_client: AsyncClient):
     create_res = await async_client.create_cluster_job(
         INPUT_FILE,
         min_cluster_size=3,
-        threshold=0.5,
+        n_neighbors=5,
```
tkmaker, line 21: I think we should keep the same defaults as UMAP and BERTopic which is I believe 15. Is a reason to go with 5 here? It might produce more local structures vs global ones.

Same for min_cluster_size. BERTopic uses a default of 10.

tkmaker (dismissed):

cohere/client.py
sanderland: The usual is to have a default of `None`, but mention the API default in the docstring, see e.g. generate.

tests/async/test_async_cluster.py
alekhya-n: Done! Setting n_neighbors to the BERTopic default of 15 and min_cluster_size to 10. This will give us 2 clusters (sizes 77, 11) and 12 outliers for this dataset.

cohere/client.py
alekhya-n: Updated docstring with default values!

cohere/client.py
```diff
@@ -597,25 +597,29 @@ def _request(self, endpoint, json=None, method="POST", stream=False) -> Any:
...
+            is_deterministic (Optional[bool], optional): Determines whether the output of the cluster job is
+            deterministic. Defaults to None.
```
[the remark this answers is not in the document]
alekhya-n: Done!

abdullahkady (approved):

abdullahkady:
this PR is no longer relevant in favor of https://github.com/cohere-ai/cohere-python/pull/207 right?

alekhya-n:
> this PR is no longer relevant in favor of #207 right?

That's right!

abdullahkady (approved):

tkmaker (approved):
LGTM!

sanderland:
I believe this was merged in with #212

PR #208, opened 2023-04-18
Fix missing CohereConnectionError import in Client

sanderland:


cohere/client.py
```diff
@@ -11,7 +11,7 @@
 from urllib3 import Retry
 
 import cohere
-from cohere.error import CohereAPIError, CohereError
+from cohere.error import CohereAPIError, CohereConnectionError, CohereError
```
abdullahkady, line 14: how was this not captured by CI linters?

abdullahkady (approved):

cohere/client.py
sanderland: They capture unused imports, not missing ones.

abdullahkady: we definitely should configure that! it's not about imports, it's for undefined variable in general:
`undefined name 'CohereConnectionError'Flake8(F821)`

sanderland: we don't have flake8 as a precommit hook though. Adding it with select errors such as this (not overlapping with formatting) seems like a good idea :)

PR #209, opened 2023-04-18
Add support for finetune sdk

raphaelcristal:
Discussion in notion: https://www.notion.so/cohereai/Finetune-SDK-3041fd7bf67344dfa9f54a55f02406fe?pvs=4
Linear: https://linear.app/cohereai/issue/FIN-26/finetune-sdk-for-platform

This PR aims to add support for finetunes to the sdk. Since we have to use the BlobheartAPI for this, the code looks slightly different than for the other endpoints.

How to use this:
```
csv_data = CsvDataset(train_file="/path/to/your.csv", delimiter=",")
client = cohere.FinetuneClient("YOUR_API_KEY")
client.api_url = "https://dev.api.os.cohere.ai"
finetune = client.create(
    name="sdk-test-123456",
    dataset=csv_data,
    finetune_type="GENERATIVE"
)
```


cohere/client.py
```diff
@@ -562,7 +566,7 @@ def _request(self, endpoint, json=None, method="POST", stream=False) -> Any:
             "Request-Source": self.request_source,
         }
 
-        url = f"{self.api_url}/{self.api_version}/{endpoint}"
+        url = f"{self.api_url}/{endpoint}"
```
raphaelcristal, line 569: Find a proper solution for this one. This breaks all the other endpoints. BlobheartAPI has no concept of api versioning.

cohere/client.py
```diff
@@ -796,3 +800,133 @@ def wait_for_bulk_embed_job(
...
+    def get_by_name(self, name: str) -> Finetune:
+        """Get a finetune by name.
+
+        Args:
+            name (str): finetune name
+        Returns:
+            Finetune: the finetune
+        """
+        json = {"name": name}
+        response = self._client._request(f"{cohere.BLOBHEART_URL}/GetFinetuneByName", method="POST", json=json)
```
raphaelcristal, line 830: find a solution to not access a private member of the client. The same goes for all he other calls.

cohere/client.py
```diff
@@ -796,3 +800,133 @@ def wait_for_bulk_embed_job(
...
+
+    def create(self, name: str, finetune_type: FINETUNE_BASE_TYPE, dataset: Dataset) -> str:
+        """Create a new finetune
+
+        Args:
+            name (str): name of your finetune, has to be unique across your organization
+            finetune_type (GENERATIVE, CONTRASTIVE, CLASSIFICATION): type of finetune
+            dataset (InMemoryDataset, CsvDataset, JsonlDataset, TextDataset): A dataset for your training. Consists of a train and optional eval file.
+        Returns:
+            str: the id of the finetune that was created
```
joon0711, line 878: Can we return the `Finetune` object instead of the id?

cohere/client.py
```diff
@@ -796,3 +800,133 @@ def wait_for_bulk_embed_job(
...
+        json = {"name": name}
+        response = self._client._request(f"{cohere.BLOBHEART_URL}/GetFinetuneByName", method="POST", json=json)
+        return Finetune.from_dict(response["finetune"])
+
+    def list(
+        self,
+        statuses: Optional[List[FINETUNE_STATUS]] = None,
+        before: Optional[datetime] = None,
+        after: Optional[datetime] = None,
+        limit: int = 50,
```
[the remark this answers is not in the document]
raphaelcristal: that's a good point, will remove.

cohere/client.py
raphaelcristal: Usually yes, but the backend return a malformed object currently. Have to fix this first, then we can return the finetune object.

cohere/finetune_dataset.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def eval_file_name(self) -> str:
+        if not self.has_eval_file():
+            raise ValueError("Dataset has no eval file")
+        return Path(self._eval_file).name
+
+    def has_eval_file(self) -> str:
+        return self._eval_file is not None
+
+    def get_train_file_contents(self) -> Iterable[bytes]:
```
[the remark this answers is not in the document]
raphaelcristal: `get_raw_train_content_iterator()`
I think there is a bit of misunderstanding how this is being used. This should not be called by a user. But instead this is called inside of the create finetune call. 
How about we name it just get_train_data ? I think from the type Iterable[bytes] it is already clear that it is an Iterator plus raw.

cohere/finetune_dataset.py
```diff
@@ -0,0 +1,171 @@
...
+        >>> InMemoryDataset([("this is a prompt", "this is a completion")])
+        >>> InMemoryDataset([("example1", "label1"), ("example2", "label1"), ("example3", "label2")])
+    """
+
+    def __init__(self, training_data: Iterable[tuple[str, str]], eval_data: Optional[Iterable[tuple[str, str]]] = None):
+        self._training_data = training_data
+        self._eval_data = eval_data
+        self._delimiter = ","
+
+    def train_file_name(self) -> str:
```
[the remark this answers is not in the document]
raphaelcristal: Private methods should only be called within the same class. So this one has to be public for the create function to be able to call it.

cohere/finetune_dataset.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def eval_file_name(self) -> str:
+        if not self.has_eval_file():
+            raise ValueError("Dataset has no eval data")
+        return "eval.csv"
+
+    def has_eval_file(self) -> str:
+        return self._eval_data is not None
+
+    def get_train_file_contents(self) -> Iterable[bytes]:
```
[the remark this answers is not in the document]
raphaelcristal: Also we shouldn't delete otherwise the code doesn't work :)

cohere/client.py
```diff
@@ -796,3 +801,131 @@ def wait_for_bulk_embed_job(
...
+        response = self._client._request(f"{cohere.BLOBHEART_URL}/CreateFinetune", method="POST", json=json)
+        return response["finetune"]["id"]
+
+    def _upload_dataset(
+        self, content: Iterable[bytes], finetune_name: str, file_name: str, type: _INTERNAL_FINETUNE_TYPE
+    ) -> str:
+        gcs = self._create_signed_url(finetune_name, file_name, type)
+        response = requests.put(gcs["url"], data=content, headers={"content-type": "text/plain"})
+        if not response.status_code == 200:
+            raise Exception(f"Unable to upload dataset: {response.text}")
```
sanderland, line 924: CohereError?

cohere/finetune_dataset.py
```diff
@@ -0,0 +1,171 @@
...
+    def file_config(self) -> FileConfig:
+        config = _empty_file_config.copy()
+        if self._separator:
+            config["separator"] = self._separator
+        return config
+
+
+class InMemoryDataset(Dataset):
+    """
+    A dataset existing in memory. You may pass a generator to avoid loading your whole dataset into memory at once.
```
sanderland, line 129: suggests you can do this with less RAM than your dataset, but you really can't as the StringIO will store it anyway?

raphaelcristal: yes but it is only for one row at a time. Or did I misunderstand how generators work?
```
for row in self._training_data:
             yield self._serialize_row(row)             
```

And then
```
def _serialize_row(self, row: tuple[str, str]) -> bytes:
         buffer = StringIO()
         writer = csv.writer(buffer, delimiter=self._delimiter)
         writer.writerow(row)
         return f"{buffer}\n".encode("utf-8")
```

sanderland: Sorry, you are right!

cohere/client.py
```diff
@@ -537,24 +547,6 @@ def rerank(
             rank.document = parsed_docs[rank.index]
         return reranking
 
-    def _check_response(self, json_response: Dict, headers: Dict, status_code: int):
```
sanderland, line 540: wont this break asyncclient

cohere/client.py
```diff
@@ -796,3 +788,181 @@ def wait_for_bulk_embed_job(
             timeout=timeout,
             interval=interval,
         )
+
+
+class FinetuneClient:
```
sanderland, line 793: separate file

cohere/__init__.py
sanderland, line 1: import FinetuneClient in this file as well

sanderland:
* document a little (at least the existence of the client) in readme
* add appropriate things in docs/ for autogenerated docs
* version 4.3.0, changelog

cohere/__init__.py
raphaelcristal: will detach the FinetuneClient from the normal Client as discussed.

cohere/responses/response.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict
+
+from cohere.error import CohereAPIError, CohereError
+from cohere.logging import logger
+
+
+def check_response(json_response: Dict, headers: Dict, status_code: int):
```
sanderland, line 7: I would throw this in utils

README.md
```diff
@@ -53,6 +53,33 @@ print('prediction: {}'.format(prediction.generations[0].text))
 
 There is also an asyncio compatible client called `cohere.AsyncClient` with an equivalent interface. Consult the [SDK Docs](https://cohere-sdk.readthedocs.io/en/latest/) for more details.
 
+If you are looking to work with finetunes you will have to instantiate a FinetuneClient
+```python
+import cohere
+from cohere.finetune_dataset import  InMemoryDataset, JsonlDataset
+
+co = cohere.Client('YOUR_API_KEY')
```
sanderland, line 61: maybe move down to where you use it as in 
# you can now call your model as usual
co = ..
co.generate

sanderland (approved):
Looks great!

mkozakov:
I'd like to discuss this more next week, not stoked about the SDK hitting `api.os.cohere.ai` as opposed to `api.cohere.ai`. This has all sorts of implications, API versioning being one of them

mkozakov:
Closing this PR until we iron out the API stuff

PR #210, opened 2023-04-18
Update embed with compressed embeddings

AmrMKayid:
Add compressed embeddings to `co.embed`

Example

```python
import cohere
co = cohere.Client(f"{API_KEY}")
response = co.embed(
  model='multilingual-22-12',
  texts=["hello", "مرحبا"],
  compress=True,
)
print('Compressed Embeddings: {}'.format(response.compressed_embeddings))
```


cohere/client.py
```diff
@@ -279,15 +279,27 @@ def _validate_chatlog_override(self, chatlog_override: List[Dict[str, str]]) ->
...
+        compress: Optional[bool] = False,
+        compression_codebook: Optional[str] = "default",
+    ) -> Embeddings:
         """Returns an Embeddings object for the provided texts. Visit https://cohere.ai/embed to learn about embeddings.
 
         Args:
             text (List[str]): A list of strings to embed.
             model (str): (Optional) The model ID to use for embedding the text.
             truncate (str): (Optional) One of NONE|START|END, defaults to END. How the API handles text longer than the maximum token length.
+            compress (bool): (Optional) Whether to compress the embeddings. When True, the compressed_embeddings will be returned as integers in the range [0, 255].
```
mkozakov, line 296: should we include a link to docs to learn more about compressed embeddings?

mkozakov (approved):

mkozakov (commented):
@lfayoux can you take a look at this to make sure we correctly add this to the beta version?

cohere/client.py
AmrMKayid: I think we will update the `embed` docs directly to include these changes and we will have a blog post published soon too that we can add to the URL that exist the docstring `https://cohere.ai/embed` cc: @mahjongmen

cohere/client_async.py
```diff
@@ -209,16 +209,41 @@ async def chat(
...
             for i in range(0, len(texts), cohere.COHERE_EMBED_BATCH_SIZE)
         ]
         responses = await asyncio.gather(*[self._request(cohere.EMBED_URL, json) for json in json_bodys])
         meta = responses[0]["meta"] if responses else None
 
-        embeddings = Embeddings([e for res in responses for e in res["embeddings"]], meta)  # concatenate results
-        return embeddings
+        return Embeddings(
+            embeddings=[e for res in responses for e in res["embeddings"]],
+            compressed_embeddings=[e for res in responses for e in res["compressed_embeddings"]] if compress else [],
```
sanderland, line 244: `else None` seems more appropriate for a parameter that doesn't apply

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cohere"
-version = "4.2.1"
+version = "4.2.2"
```
sanderland, line 3: I think 4.3 is appropriate for a new endpoint

sanderland (approved):
two small suggestions but otherwise looks good

cohere/client_async.py
AmrMKayid: addressed!

pyproject.toml
AmrMKayid: updated

cohere/client.py
AmrMKayid: @mkozakov created this issue here https://github.com/cohere-ai/cohere-python/issues/216 and I'll update this once we have the docs ready on the website, resolving this for now

PR #219, opened 2023-05-03
Update Compressed embeddings indices names

AmrMKayid:
Updating the compressed embeddings indices names to include the dimension instead of the compression ratios

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -60,19 +60,21 @@ python-versions = ">=3.5"
 
 [[package]]
 name = "attrs"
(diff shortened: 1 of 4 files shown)


AmrMKayid:
> Could you please fix the failing tests in embeddings?

The test failure happens in generate tests

sanderland (approved):
lgtm
#221 is set to merge into your branch and should make tests pass, feel free to merge everything like that

sanderland (approved):

PR #220, opened 2023-05-03
Chat - Add chat_history

lusmoura:
### What this PR does
- Add support to chat_history
- Update the default `max_tokens` to None
- Closes CNV-574

### DIFF
--- a/cohere/client_async.py
+++ b/cohere/client_async.py
@@ -191,11 +191,6 @@ class AsyncClient(Client):
         if chat_history is not None:
             self._validate_chat_history(chat_history)
 
-        if max_tokens is not None:
-            logger.warning(
-                "The 'max_tokens' parameter is deprecated and will be removed in a future version of this function.",
-            )
-
         json_body = {
             "query": query,
             "conversation_id": conversation_id,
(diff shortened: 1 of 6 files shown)


sanderland (approved):

cohere/client_async.py
```diff
@@ -178,15 +178,19 @@ async def chat(
...
         chatlog_override: List[Dict[str, str]] = None,
+        chat_history: List[Dict[str, str]] = None,
         preamble_override: str = None,
         user_name: str = None,
         temperature: float = 0.8,
-        max_tokens: int = 200,
+        max_tokens: int = None,
         stream: bool = False,
     ) -> Union[AsyncChat, StreamingChat]:
         if chatlog_override is not None:
```
abdullahkady, line 188: shouldn't we add the same docstring? I know the whole setup of duplication is not optimal, but it is what we have atm.

abdullahkady (approved):

cohere/client_async.py
lusmoura: I think the docstring is duplicated when rendering [the documentation](https://cohere-sdk.readthedocs.io/en/latest/cohere.html#cohere.client_async.AsyncClient.chat)

abdullahkady: @lusmoura I was thinking for code intellisense and editor hints. It's ok 👌

PR #232, opened 2023-05-18
Support model param in de/tokenize

abdullahkady:


tests/sync/test_tokenize.py
```diff
@@ -8,6 +9,17 @@
 
 
 class TestTokenize(unittest.TestCase):
+    @pytest.mark.skipif(True, reason="until this is available globally in the API")
```
harry-cohere, line 12: I believe the sync tests don't use pytest. Instead you can skip the test using the unittest decorators: https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

harry-cohere (changes requested):

tests/sync/test_tokenize.py
abdullahkady: @harry-cohere we still use pytest as our runner, so the decorators will take effect normally:
https://github.com/cohere-ai/cohere-python/blob/main/.github/workflows/test.yaml#L29-L32

tests/async/test_async_tokenize.py
```diff
@@ -25,14 +38,27 @@ async def test_invalid_text(async_client):
...
-    detokenizeds = await async_client.batch_detokenize([[10104, 12221, 974, 514, 34]] * 3)
+    detokenizeds = await async_client.batch_detokenize([[10104, 12221, 974, 514, 34]] * 3, model="xlarge")
     assert detokenizeds == [Detokenization("detokenize me!")] * 3
 
     detokenized = await async_client.detokenize([])
     assert detokenized == Detokenization("")
+
+
+@pytest.mark.asyncio
+@pytest.mark.skipif(True, reason="until this is available globally in the API")
```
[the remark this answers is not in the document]
abdullahkady: just fyi .. these comments shouldn't be relevant cuz the plan is to remove these skip lines within the PR before merging anyway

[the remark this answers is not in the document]
abdullahkady: I'm just using it now to run CI 😅

harry-cohere (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.6.2
```
sanderland, line 3: would do 4.7, this is not a bugfix but a new feature

cohere/client.py
```diff
@@ -432,53 +432,61 @@ def summarize(
...
-    def tokenize(self, text: str) -> Tokens:
+    def tokenize(self, text: str, model: Optional[str] = None) -> Tokens:
         """Returns a Tokens object of the provided text, see https://docs.cohere.ai/reference/tokenize for advanced usage.
 
         Args:
             text (str): Text to summarize.
+            model (str): An optional model name that will ensure that the tokenization uses the tokenizer used by that model, which can be critical for counting tokens properly.
         """
         json_body = {"text": text}
+        if model is not None:
```
sanderland, line 458: does the API require this? we usually just pass the None on

tests/async/test_async_tokenize.py
```diff
@@ -25,14 +38,27 @@ async def test_invalid_text(async_client):
 
 @pytest.mark.asyncio
 async def test_detokenize(async_client):
-    detokenized = await async_client.detokenize([10104, 12221, 974, 514, 34])
+    detokenized = await async_client.detokenize([10104, 12221, 974, 514, 34], model="xlarge")
```
sanderland, line 41: we call this 'base' now right

sanderland (approved):

mkozakov (approved):

PR #245, opened 2023-06-07
Add chat params - p, k and logit bias

lusmoura:
- Adds new params to the SDK
- Closes CNV-553
- Closes CNV-554
- Closes CNV-555

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -192,6 +192,9 @@ class Client:
         temperature: float = 0.8,
         max_tokens: int = None,
         stream: bool = False,
+        p: float = None,
+        k: float = None,
+        logit_bias: Dict[int, float] = {},
     ) -> Union[Chat, StreamingChat]:
         """Returns a Chat object with the query reply.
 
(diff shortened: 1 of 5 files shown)


cohere/client.py
```diff
@@ -192,6 +192,9 @@ def chat(
         temperature: float = 0.8,
         max_tokens: int = None,
         stream: bool = False,
+        p: float = None,
+        k: float = None,
+        logit_bias: Dict[int, float] = {},
```
sanderland, line 197: [empty dicts are a dangerous default value](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python). 
Typically you just use "None" here and have the API just handle that the same as {}

cohere/client.py
```diff
@@ -284,6 +295,16 @@ def _validate_chat_history(self, chat_history: List[Dict[str, str]]) -> None:
             if not isinstance(entry["user_name"], str) or not isinstance(entry["text"], str):
                 raise CohereError(message="both user_name and text must be strings in chat_history.")
 
+    def _validate_chat_params(self, logit_bias: Dict[int, float]) -> None:
```
sanderland, line 298: I think we just leave this up to the API. also a float,int check can easily fail with numpy types or such, and you end up with rather complicated checks (e.g. [numbers.Number](https://docs.python.org/library/numbers.html#numbers.Number))

sanderland (changes requested):
Few small changes suggested

cohere/client_async.py
```diff
@@ -196,6 +196,9 @@ async def chat(
         temperature: float = 0.8,
         max_tokens: int = None,
         stream: bool = False,
+        p: float = None,
+        k: float = None,
```
harry-cohere, line 200:
```suggestion
        p: Optional[float] = None,
        k: Optional[float] = None,
```

sanderland (approved):

PR #250, opened 2023-06-15
Support augmented documents and search queries in chat (non-streaming)

BeatrixCohere:
Closes PTS-4548

Adds functionality to chat to support augmented documents and generating search queries in chat.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -29,7 +29,7 @@ from cohere.responses import (
     Tokens,
 )
 from cohere.responses.bulk_embed import BulkEmbedJob, CreateBulkEmbedJobResponse
-from cohere.responses.chat import Chat, StreamingChat
+from cohere.responses.chat import Chat, Mode, StreamingChat
 from cohere.responses.classify import Example as ClassifyExample
 from cohere.responses.classify import LabelPrediction
 from cohere.responses.cluster import ClusterJobResult, CreateClusterJobResponse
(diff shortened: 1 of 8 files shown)


cohere/client.py
```diff
@@ -269,6 +276,16 @@ def chat(
                 >>>     return_prompt=True)
                 >>> print(res.text)
                 >>> print(res.prompt)
+            Query generation example:
+                >>> res = co.chat(query="What are the tallest penguins?", mode="search_query_generation")
```
mkozakov, line 280: i think we want to use the enum values for mode, not strings right? this applies to the tests as well

cohere/responses/chat.py
```diff
@@ -1,25 +1,31 @@
 import json
+from enum import Enum
 from typing import Any, Dict, Generator, List, NamedTuple, Optional
 
 import requests
 
 from cohere.responses.base import CohereObject
 
+Mode = Enum("Mode", ["chat", "search_query_generation", "augmented_generation"])
```
mkozakov, line 9: i think enum values are normally all caps

mkozakov (approved):

cohere/client.py
sanderland: that can be somewhat annoying to use though, having to import the enum object from somewhere, and other endpoints just take strings.

cohere/responses/chat.py
sanderland: Have never really used Enum, but after testing around a bit, this doesn't quite seem what you want?
```
Mode = Enum("Mode", ["chat", "search_query_generation", "augmented_generation"])
Mode.chat=='chat' # False
Mode.chat.value == 1 # True
```

cohere/client.py
BeatrixCohere: This should work with both importing mode and with the string

mkozakov:
Do you know why tests are still failing?

mkozakov (approved):

PR #248, opened 2023-06-14
Chat - Update request params

lusmoura:
## Description
- Add support for `message`
- Add deprecation warning for `query`
- Add support for `chat_history` `message`
- Add deprecation warning for `chat_history` `text`
- Deprecate `chatlog_override`
- Updates unit tests

## Linear
- Closes CNV-736
- Closes CNV-698
- Closes CNV-705

## Examples
```
>>> co.chat(message="hey")
cohere.Chat {
        id: 36ef1c39-81dc-4625-beaf-55eb364e6fbc
        response_id: 36ef1c39-81dc-4625-beaf-55eb364e6fbc
        generation_id: 284f785e-0f44-4ef8-bfc8-90afc01de37f
        query: hey
        message: hey
        text: Hello! How may I assist you today?
        conversation_id: c2f2d3f7-1c0d-4afd-8f23-63d34b1dcdae
        prompt: None
        chatlog: None
        preamble: None
        client: <cohere.client.Client object at 0x1028bb880>
        token_count: {'prompt_tokens': 63, 'response_tokens': 9, 'total_tokens': 72}
        meta: {'api_version': {'version': '1'}}
}
```

```
>>> co.chat(query="hey")
The 'query' parameter is deprecated and will be removed in a future version of this function. Use 'message' instead.
cohere.Chat {
        id: b9e28a5b-f1eb-4793-9fee-a0d10d2044e3
        response_id: b9e28a5b-f1eb-4793-9fee-a0d10d2044e3
        generation_id: c4255ba2-6fe6-4199-bff6-b33bfc7335b1
        query: hey
        message: hey
        text: Hi! How can I help you today?
        conversation_id: c347f5c7-6bef-46ab-8751-6fb3c099cd75
        prompt: None
        chatlog: None
        preamble: None
        client: <cohere.client.Client object at 0x1028bb880>
        token_count: {'prompt_tokens': 63, 'response_tokens': 9, 'total_tokens': 72}
        meta: {'api_version': {'version': '1'}}
}
```

```
>>> res = co.chat(message="hey", chat_history=[{"user_name": "User", "message": "hey"}])
```

```
>>> res = co.chat(message="hey", chat_history=[{"user_name": "User", "text": "hey"}])
The chat_history 'text' key is deprecated and will be removed in a future version of this function. Use 'message' instead.
```

PR #253, opened 2023-06-21
Rename query param in chat

mkozakov:
Rename the `query` param of the Chat API to `message`, to match the change that happened on the API side


lusmoura (commented):
I think this is the same as #248 👀 We're also updating the `chatlog_history` `text` to `message` ([here](https://github.com/cohere-ai/blobheart/pull/9794)), so I didn't publish the update here to have all changes in a single PR

PR #257, opened 2023-07-10
Model metrics tests

raphaelcristal:
Add tests for model metric endpoint

### DIFF
--- a/cohere/responses/custom_model.py
+++ b/cohere/responses/custom_model.py
@@ -81,6 +81,9 @@ class ModelMetric(CohereObject):
     step_num: int
     loss: float
     accuracy: float
+    f1: float
+    precision: float
+    recall: float
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> "ModelMetric":
@@ -90,8 +93,12 @@ class ModelMetric(CohereObject):
             created_at=_parse_date(dt_concat),
             step_num=data["step_num"],
             loss=data["loss"],
-            accuracy=data["accuracy"]
+            accuracy=data["accuracy"],
+            f1=data["f1"],
+            precision=data["precision"],
+            recall=data["recall"],
         )
+
 
 def _parse_date(datetime_string: str) -> datetime:
     return datetime.strptime(datetime_string, "%Y-%m-%dT%H:%M:%S.%f%z")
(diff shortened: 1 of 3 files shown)


cohere/responses/custom_model.py
```diff
@@ -81,6 +81,9 @@ class ModelMetric(CohereObject):
     step_num: int
     loss: float
     accuracy: float
+    f1: float
```
BeatrixCohere, line 84: Are these optional?

raphaelcristal: Looking at blob heart it doesn't seem like it:
https://github.com/cohere-ai/blobheart/blob/9debecc7a68d9311662d9b1e183ba09504e7bec6/proto/blobheart.gen.go#L1834-L1833

tests/test_custom_model.py
```diff
@@ -0,0 +1,26 @@
+from datetime import datetime, timezone
```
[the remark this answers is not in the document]
raphaelcristal: Not sure what else to test there. It basically just makes a http call and parses the response. Open to suggestions though.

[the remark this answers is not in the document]
BeatrixCohere: It would test the HTTP call. We have these tests for most of the HTTP calls. For example if the response came back without one of those parameters this test would catch that.

cohere/responses/custom_model.py
BeatrixCohere: It has omitempty for all these fields what happens when no field is returned?

raphaelcristal: Is this possible though if the datatype is float32 and not *float32? Wouldn't it be 0 but not missing?

raphaelcristal: Or wait does it drop the values if they are 0? Let me check that.

raphaelcristal: >The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

Okay my bad, though this was only true for nil.

cohere/responses/custom_model.py
```diff
@@ -79,8 +79,11 @@ def from_dict(cls, data: Dict[str, Any]) -> "CustomModel":
 class ModelMetric(CohereObject):
     created_at: datetime
     step_num: int
-    loss: float
-    accuracy: float
+    loss: Optional[float] = None
+    accuracy: Optional[float] = None
+    f1: Optional[float] = None
+    precision: Optional[float] = None
+    recall: Optional[float] = None
```
innainu, line 86: ah sry, I went off a curl of a generative request, totally forgot about this. agree w the optional!

PR #276, opened 2023-08-04
Alexandre/add base model option

alex-matton:
Add a new `base_model` option for reranker finetuning to decide which base model to finetune from: `english` or `multilingual`.
 For the other models (which don't have options), make sure that this variable is not filled.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -1009,25 +1009,12 @@ class Client:
         """
         internal_custom_model_type = CUSTOM_MODEL_PRODUCT_MAPPING[model_type]
 
-        # Figuring out base model
-        if internal_custom_model_type in ["GENERATIVE", "CLASSIFICATION"]:
-            assert base_model is None, "base_model has to be None for generative and classification models"
-            internal_base_model = "medium"
-        elif internal_custom_model_type == "RERANK":
-            internal_base_model = base_model or "english"
-            assert internal_base_model in [
-                "english",
-                "multilingual",
-            ], "base_model has to be `english` or `multilingual`"
-        else:
-            raise ValueError(f"Unsupported model_type: {internal_custom_model_type}")
-
         json = {
             "name": name,
             "settings": {
                 "trainFiles": [],
                 "evalFiles": [],
-                "baseModel": internal_base_model,
+                "baseModel": base_model,
                 "finetuneType": internal_custom_model_type,
             },
         }
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -972,6 +972,7 @@ class Client:
         name: str,
         model_type: CUSTOM_MODEL_TYPE,
         dataset: CustomModelDataset,
+        base_model: Optional[str] = None,
         hyperparameters: Optional[HyperParametersInput] = None,
     ) -> CustomModel:
         """Create a new custom model
@@ -980,6 +981,11 @@ class Client:
             name (str): name of your custom model, has to be unique across your organization
             model_type (GENERATIVE, CLASSIFY, RERANK): type of custom model
             dataset (InMemoryDataset, CsvDataset, JsonlDataset, TextDataset): A dataset for your training. Consists of a train and optional eval file.
+            base_model (str): base model to use for your custom model.
+                For generative and classify models, `base_model` has to be None (no option available for now)
+                For rerank models, you can choose between `english` and `multilingual`. Defaults to `english` if not specified.
+                    The English model is better for English, while the multilingual model should be picked if a non-negligible part of queries/documents
+                    will be in other languages
             hyperparameters (HyperParametersInput): adjust hyperparameters for your custom model. Only for generative custom models.
         Returns:
             str: the id of the custom model that was created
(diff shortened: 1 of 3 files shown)


cohere/client.py
```diff
@@ -1002,12 +1008,26 @@ def create_custom_model(
 
         """
         internal_custom_model_type = CUSTOM_MODEL_PRODUCT_MAPPING[model_type]
+
+        # Figuring out base model
+        if internal_custom_model_type in ["GENERATIVE", "CLASSIFICATION"]:
+            assert base_model is None, "base_model has to be None for generative and classification models"
```
mkozakov, line 1014: this validation should happen on the backend

cohere/client.py
```diff
@@ -1002,12 +1008,26 @@ def create_custom_model(
...
+            internal_base_model = "medium"
+        elif internal_custom_model_type == "RERANK":
+            internal_base_model = base_model or "english"
+            assert internal_base_model in [
```
mkozakov, line 1018: same here

cohere/client.py
```diff
@@ -1002,12 +1008,26 @@ def create_custom_model(
 
         """
         internal_custom_model_type = CUSTOM_MODEL_PRODUCT_MAPPING[model_type]
+
+        # Figuring out base model
+        if internal_custom_model_type in ["GENERATIVE", "CLASSIFICATION"]:
+            assert base_model is None, "base_model has to be None for generative and classification models"
+            internal_base_model = "medium"
+        elif internal_custom_model_type == "RERANK":
+            internal_base_model = base_model or "english"
```
mkozakov, line 1017: the default should be set on the backend

cohere/responses/custom_model.py
```diff
@@ -104,6 +106,7 @@ def from_dict(cls, data: Dict[str, Any]) -> "CustomModel":
             model_type=REVERSE_CUSTOM_MODEL_PRODUCT_MAPPING[data["settings"]["finetuneType"]],
             created_at=_parse_date(data["created_at"]),
             completed_at=_parse_date(data["completed_at"]) if "completed_at" in data else None,
+            base_model=data["settings"]["baseModel"],
```
mkozakov, line 109: is `baseModel` correct? shouldn't it be `base_model` https://github.com/cohere-ai/blobheart/blob/main/go/pkg/data/finetune.go#L33

cohere/client.py
alex-matton: Oh we had a discussion with Inna and Raphael, and we thought that it made more sense to do an internal mapping here like it's done for `internal_custom_model_type` with `internal_custom_model_type = CUSTOM_MODEL_PRODUCT_MAPPING[model_type]`

The other solution without the mapping would be to:
- use `base_model: str = "medium"` as default argument of the function
- this would force the user to specify `base_model="english"` or `base_model="multilingual"` for reranker.

What do you think is the best solution?

cohere/responses/custom_model.py
alex-matton: Yes I tried it again, and it's `baseModel` and not `base_model`. This is consistent with the line 106 above with `data["settings"]["finetuneType"]` instead of `data["settings"]["finetune_type"]`

cohere/client.py
mkozakov: - for non-rerank finetuning, we want to continue supporting `None` as a value for the base_model. That allows us to change the default value on the backend.

- for rerank finetuning, the backend should throw an error "base_model cannot be empty for finetune type "rerank". Must be one of "english" or "multilingual"

no need to keep a map of defaults on the client

cohere/responses/custom_model.py
mkozakov: weird. do you know where these values are coming from? we should try to make it consistent if possible, otherwise i anticipate bugs

alex-matton: The class to look at is that one: https://github.com/cohere-ai/blobheart/blob/alexandre/add_basemodel_rerank/proto/blobheart.gen.go#L1546

mkozakov: ooof. this is very inconsistent. ok let's keep it as is for now, and also bring this up to the OS team

cohere/client.py
alex-matton: For non-rerank finetuning, the base_model value has always been "medium": https://github.com/cohere-ai/cohere-python/blob/d21c04bab8e2de3960012e928c3ff7dcd42b9869/cohere/client.py#L1120
-> that's why I'm thinking of setting "medium" as the default value (ensures no breaking change) if we don't keep a map of defaults on the client

mkozakov: just like for all other apis, we should not be setting any defaults in the client, so just remove that "medium" setting from the SDK and make sure that the backend sets the correct value

alex-matton: @raphaelcristal tagging you here as this is probably something you own. Do you think it's possible to do what @mkozakov is suggesting above? I.e. remove the `"baseModel":"medium"` default and instead add defaults to the backend?

raphaelcristal: Yes what Michael says is correct. Best to keep all validation on the backend, so we don't have to deal with deprecations in the sdk. Sorry if I didn't explain this properly when we talked.

mkozakov (approved):
Thank you for making the changes!

sanderland (approved):

PR #277, opened 2023-08-10
implement loglikelihood endpoint

sanderland:
implements co.loglikelihood
required for evals

also resolves #264 and #269 by poetry update


tests/sync/test_llh.py
```diff
@@ -0,0 +1,14 @@
+from utils import get_api_key
```
abdullahkady, line 1: I'd suggest not using llh for file name. it's usually easier to have the same naming between code/tests

cohere/client.py
```diff
@@ -104,6 +105,25 @@ def check_api_key(self) -> Dict[str, bool]:
...
+        completion: Optional[str] = None,
+        model: Optional[str] = None,
+    ) -> LogLikelihoods:
+        """Calculates the token log-likelihood for a provided prompt and completion.
+        Using this endpoint instead of co.generate with max_tokens=0 will guarantee that any required tokens such as <EOP_TOKEN>
+        are correctly inserted, and makes it easier to retrieve only the completion log-likelihood.
+
+        Args:
+            prompt (str): The prompt
+            completion (str): The completion
```
abdullahkady, line 120: should be optional

cohere/responses/loglikelihood.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class LogLikelihoods:
+    @staticmethod
+    def token_list_from_dict(token_list: List[Dict]):
+        return [TokenLogLikelihood(**token) for token in token_list]
+
+    def __init__(self, prompt_tokens: List[TokenLogLikelihood], completion_tokens: List[TokenLogLikelihood]):
+        self.prompt_tokens = self.token_list_from_dict(prompt_tokens)
+        self.completion_tokens = self.token_list_from_dict(completion_tokens)
```
abdullahkady, line 13: I think this would error out if you provide prompt only. trying to iterate over `None` for completion tokens

tests/sync/test_llh.py
```diff
@@ -0,0 +1,14 @@
...
+
+import cohere
+from cohere.responses import LogLikelihoods
+
+TEST_MODEL = "command-light"
+API_KEY = get_api_key()
+co = cohere.Client(API_KEY)
+
+
+async def test_basic_llh():
```
abdullahkady, line 11: this is sync tests. I don't think we use async (and we seem to use `unittest.TestCase`, not just functions)

abdullahkady (changes requested):

cohere/client.py
```diff
@@ -104,6 +105,25 @@ def check_api_key(self) -> Dict[str, bool]:
         """
         return {"valid": is_api_key_valid(self.api_key)}
 
+    def loglikelihood(
+        self,
+        prompt: Optional[str] = None,
+        completion: Optional[str] = None,
+        model: Optional[str] = None,
```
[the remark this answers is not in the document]
sanderland: I thought we were keeping those options out of release for now

cohere/responses/loglikelihood.py
sanderland: nice catch

sanderland: added test

.github/workflows/test.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.9", "3.11"] # assume min and max are sufficient
+        python-version: ["3.7", "3.11"] # assume min and max are sufficient
```
lfayoux, line 11: can we keep 3.9 in the test matrix? it was failing before

lfayoux (approved):
lgtm

PR #279, opened 2023-08-14
fix dataset list key error

lfayoux:
if there are no datasets to list return an empty list


cohere/client.py
```diff
@@ -765,10 +765,14 @@ def list_datasets(self, dataset_type: str = None, limit: int = None, offset: int
...
-        return [
-            Dataset.from_dict({"meta": response.get("meta"), **r}, wait_fn=self.wait_for_dataset)
-            for r in response["datasets"]
-        ]
+        return (
+            [
+                Dataset.from_dict({"meta": response.get("meta"), **r}, wait_fn=self.wait_for_dataset)
+                for r in response["datasets"]
+            ]
+            if "datasets" in response
```
sanderland, line 773: I would use `for r in response.get('datasets',[])` or `for r in (response.get('datasets') or [])` but that's just a suggestion

sanderland (approved):

harry-cohere (approved):

PR #281, opened 2023-08-15
fix llh return object and add jupyter visualization

sanderland:
No need for a version or changelog entry since this is largely cosmetic.


cohere/responses/loglikelihood.py
```diff
@@ -13,3 +15,21 @@ def token_list_from_dict(token_list: Optional[List[Dict]]):
     def __init__(self, prompt_tokens: List[TokenLogLikelihood], completion_tokens: List[TokenLogLikelihood]):
         self.prompt_tokens = self.token_list_from_dict(prompt_tokens)
         self.completion_tokens = self.token_list_from_dict(completion_tokens)
+
+    def visualize(self, **kwargs):
+        import pandas as pd
```
mkozakov, line 20: We should handle gracefully if user doesn't have pandas installed

sanderland: We do https://github.com/cohere-ai/cohere-python/blob/main/cohere/responses/base.py#L83

lfayoux (approved):

PR #287, opened 2023-08-21
Support event type when streaming chat. Remove deprecated parameters. 

BeatrixCohere:
- Supports event types for chat 
- Cleans up parameters for chat 
- Removes deprecated query parameters

### DIFF
--- a/cohere/responses/chat.py
+++ b/cohere/responses/chat.py
@@ -91,12 +91,12 @@ class StreamStart(StreamResponse):
     def __init__(
         self,
         generation_id: str,
-        conversation_id: str,
+        conversation_id: Optional[str],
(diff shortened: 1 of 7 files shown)


cohere/responses/chat.py
```diff
@@ -25,10 +25,9 @@ def __init__(
         super().__init__(**kwargs)
         self.response_id = response_id
         self.generation_id = generation_id
-        self.query = message  # to be deprecated
         self.message = message
         self.text = text
-        self.conversation_id = conversation_id
+        self.conversation_id = conversation_id  # optional
```
harry-cohere, line 30: Should be type hinted as Optional?

PR #289, opened 2023-08-23
Require poetry-core instead of poetry

tjni:
In [nixpkgs](https://github.com/NixOS/nixpkgs), we build Python packages using [`build`](https://pypa-build.readthedocs.io/en/stable/) instead of poetry, and so we only need to use the more lightweight poetry-core as the PEP 517 build backend.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,5 +37,5 @@ Sphinx = "^5.0"
 sphinx-rtd-theme = "^1.2.0"
 
 [build-system]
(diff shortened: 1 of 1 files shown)


tjni:
Oh, I'm sorry I didn't realize a CLA is involved. Personally, I don't read or sign CLAs, so I'm going to close this PR. Of course, this change is not copyrightable so please feel free to reproduce it if you'd like.

PR #292, opened 2023-08-30
Add search query and citations to the chat python SDK 

BeatrixCohere:
Close: PTS-4743

This adds:
- Search query only parameter
- Chat with documents 
- Chat with connectors
- Citation quality parameter

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -270,14 +270,24 @@ class Client:
 
             user_name (str): (Optional) A string to override the username.
 
-            // TODO
-
-            search_queries_only: Optional[bool] = None,
-
-            documents: Optional[List[Dict[str, str]]] = None,
-            connectors: Optional[List[Dict[str, str]]] = None,
-
-            citation_quality: Optional[str] = None,
+            search_queries_only (bool) : (Optional) When true, the response will only contain a list of generated search queries, but no search will take place, and no reply from the model to the user's message will be generated.
+            documents (List[Dict[str, str]]): (Optional) Documents to use to generate grounded response with citations. Example:
+                documents=[
+                    {
+                        "id": "national_geographic_everest",
+                        "title": "Height of Mount Everest",
+                        "snippet": "The height of Mount Everest is 29,035 feet",
+                        "url": "https://education.nationalgeographic.org/resource/mount-everest/",
+                    },
+                    {
+                        "id": "national_geographic_mariana",
+                        "title": "Depth of the Mariana Trench",
+                        "snippet": "The depth of the Mariana Trench is 36,070 feet",
+                        "url": "https://www.nationalgeographic.org/activity/mariana-trench-deepest-place-earth",
+                    },
+                ],
+            connectors (List[Dict[str, str]]): (Optional) When specified, the model's reply will be enriched with information found by quering each of the connectors (RAG). Example: connectors=[{"id": "web-search"}]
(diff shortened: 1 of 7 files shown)


cohere/client.py
```diff
@@ -265,6 +269,25 @@ def chat(
...
+                        "url": "https://education.nationalgeographic.org/resource/mount-everest/",
+                    },
+                    {
+                        "id": "national_geographic_mariana",
+                        "title": "Depth of the Mariana Trench",
+                        "snippet": "The depth of the Mariana Trench is 36,070 feet",
+                        "url": "https://www.nationalgeographic.org/activity/mariana-trench-deepest-place-earth",
+                    },
+                ],
+            connectors (List[Dict[str, str]]): (Optional) When specified, the model's reply will be enriched with information found by quering each of the connectors (RAG). Example: connectors=[{"id": "web-search"}]
```
sanderland, line 289: For connectors, how would someone find the allowed values?

cohere/client.py
```diff
@@ -314,7 +370,12 @@ def chat(
             "p": p,
             "k": k,
             "logit_bias": logit_bias,
+            "search_queries_only": search_queries_only,
+            "documents": documents,
+            "connectors": connectors,
         }
+        if citation_quality is not None:
+            json_body["citation_quality"] = citation_quality
```
sanderland, line 378: does the API not like None as default here?

cohere/responses/chat.py
```diff
@@ -152,6 +216,11 @@ def _make_response_item(self, index, line) -> Any:
             self.chatlog = response.get("chatlog")
             self.token_count = response.get("token_count")
             self.meta = response.get("meta")
+            self.is_search_required = (response.get("is_search_required"),)  # optional
```
sanderland, line 219: why the tuples here?

sanderland (commented):
Looks great. Mainly concerned about the single item tuples being wrong.

harry-cohere (approved):

cohere/client.py
BeatrixCohere: For the short term, this will be documented. For the longer term, there will be a call to the API to list connectors.

sanderland (approved):

cohere/responses/chat.py
```diff
@@ -75,16 +91,27 @@ async def respond(self, response: str, max_tokens: int = None) -> "AsyncChat":
         )
 
 
+class Events(str, Enum):
```
mkozakov, line 94: Should be singular `StreamEvent`?

cohere/responses/chat.py
```diff
@@ -75,16 +91,27 @@ async def respond(self, response: str, max_tokens: int = None) -> "AsyncChat":
...
+    TEXT_GENERATION = "text-generation"
+    CITATION_GENERATION = "citation-generation"
+    STREAM_END = "stream-end"
+
+
 class StreamResponse(CohereObject):
     def __init__(
         self,
         is_finished: bool,
+        event_type: str,
```
mkozakov, line 107:
```suggestion
        event_type: typing.Union([StreamEvent, str])
```

cohere/client.py
```diff
@@ -296,6 +319,39 @@ def chat(
                 >>>     return_prompt=True)
                 >>> print(res.text)
                 >>> print(res.prompt)
+            Generate search queries for fetching documents to use in chat:
+                >>> res = co.chat(
+                >>>     "What is the height of Mount Everest?",
+                >>>      search_queries_only=True)
```
mkozakov, line 325: up to you, but this is the most niche use case, so imo probably should be at the bottom.

PR #306, opened 2023-09-15
fix attribute error in CohereAPIError

sanderland:
The aiohttp return object has no status_code, but status.


andyt-cohere:
Thank you so much Sander!

PR #308, opened 2023-09-19
Dataset: add validation warnings

raphaelcristal:
Expose fields from the backend.

### DIFF
--- a/tests/sync/test_chat.py
+++ b/tests/sync/test_chat.py
@@ -1,6 +1,6 @@
 import unittest
-import pytest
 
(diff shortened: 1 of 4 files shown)


tests/sync/test_chat.py
```diff
@@ -1,6 +1,6 @@
 import unittest
-import pytest
 
+import pytest
```
raphaelcristal, line 3: not sure why this was moved. was done by the recommit

lfayoux (approved):

PR #314, opened 2023-09-29
Add custom ChatHistoryEntry class

mkozakov:
`chat_history` elements are structured, however today we accept a str,str dictionary. Having a dedicated class makes it easier for the user to figure out how to use the API.


tests/sync/test_chat.py
```diff
@@ -144,8 +145,8 @@ def test_chat_history(self):
         prediction = co.chat(
             "Who are you?",
             chat_history=[
-                {"role": "User", "message": "Hey!"},
-                {"role": "Chatbot", "message": "Hey! How can I help you?"},
+                ChatHistoryEntry(message="Hey!", role=Role.USER),
```
sanderland, line 148: the old format is still valid, right? Can we keep this tested?

PR #324, opened 2023-10-13
Add support for multilabel classification

alex-matton:
- multilabel classification adds two new fields: `predictions` and `confidences`. Add support for these fields.
- mutlilabel classification doesn't return `prediction` and `confidence` fields. Mark these fields as optional.
- `prediction` and `confidence` fields will eventually be removed. Add "to be deprecated" comments.

Important note: this PR does not introduce any breaking change.

### DIFF
--- a/cohere/responses/classify.py
+++ b/cohere/responses/classify.py
@@ -31,9 +31,9 @@ class Classification(CohereObject):
 
         if self._prediction is None or self._confidence is None:
             if self._prediction is not None or self._confidence is not None:
-                raise ValueError("Cannot have one of prediction and confidence be None and not the other one")
+                raise ValueError("Cannot have one of `prediction` and `confidence` be None and not the other one")
             if self.predictions is None or self.confidences is None:
-                raise ValueError("Cannot have predictions or confidences be None if prediction is None")
+                raise ValueError("Cannot have `predictions` or `confidences` be None if `prediction` is None")
 
     def __repr__(self) -> str:
         if self._prediction is not None:
--- a/cohere/responses/classify.py
+++ b/cohere/responses/classify.py
@@ -8,19 +8,34 @@ Example = NamedTuple("Example", [("text", str), ("label", str)])
 
 class Classification(CohereObject):
     def __init__(
-        self, input: str, prediction: str, confidence: float, labels: Dict[str, LabelPrediction], *args, **kwargs
+        self,
+        input: str,
+        predictions: Optional[List[str]],
+        confidences: Optional[List[float]],
+        prediction: Optional[str],
+        confidence: Optional[float],
+        labels: Dict[str, LabelPrediction],
+        *args,
+        **kwargs,
     ) -> None:
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


cohere/responses/classify.py
```diff
@@ -8,19 +8,34 @@
...
+        confidence: Optional[float],
+        labels: Dict[str, LabelPrediction],
+        *args,
+        **kwargs,
     ) -> None:
         super().__init__(*args, **kwargs)
         self.input = input
-        self.prediction = prediction
-        self.confidence = confidence
+        self.prediction = prediction  # to be deprecated
```
sanderland, line 23: how do you expect to deprecate this field in the future?

alex-matton: @lfayoux You're probably better positioned to answer this question than me. I'm not sure what's the standard procedure for field deprecation.

lfayoux: This will be deprecated when we do an api version change. Not sure if we still want to support this field in the future

cohere/responses/classify.py
```diff
@@ -8,19 +8,34 @@
...
+        self.confidences = confidences
         self.labels = labels
 
     def __repr__(self) -> str:
-        prediction = self.prediction
-        confidence = self.confidence
-        labels = self.labels
-        return f'Classification<prediction: "{prediction}", confidence: {confidence}, labels: {labels}>'
+        if self.prediction is not None:
+            assert self.confidence is not None
```
raphaelcristal, line 31: I am not sure why you have an assert in the __repr__ function? What are you trying to achieve here?

cohere/responses/classify.py
alex-matton: Took that into account

alex-matton:
@sanderland I implemented all your requested changes 👍. For the tests it is not possible to use a real co.classify() call as it would require to finetune a model first (which takes ~30min). Instead, I mocked the `_request()` call and tried all the possible format outputs.

raphaelcristal (approved):

cohere/responses/classify.py
```diff
@@ -8,19 +9,50 @@
...
+        self._prediction = prediction  # to be deprecated
+        self._confidence = confidence  # to be deprecated
+        self.predictions = predictions
+        self.confidences = confidences
         self.labels = labels
+        self.classification_type = classification_type
+
+        if self._prediction is None or self._confidence is None:
+            if self._prediction is not None or self._confidence is not None:
+                raise ValueError("Cannot have one of prediction and confidence be None and not the other one")
```
mkozakov, line 34:
```suggestion
                raise ValueError("Cannot have one of `prediction` and `confidence` be `None` and not the other one")
```

cohere/responses/classify.py
mkozakov: logging warnings for now is ok. when we deprecate we will need to bump the major SDK version

sanderland (approved):

lfayoux (approved):

PR #329, opened 2023-10-18
Fix finetune test

harry-cohere:
Fixes a broken test but keeps a similar assertion.


lfayoux (approved):

raphaelcristal:
The real problem is the endpoint no the test unfortunately.

PR #341, opened 2023-11-12
Fix: updated co.generate docstring

fparadas:
The function docstring was saying that the `truncate` argument was `"END"` on default, but actually it is `None`.

This PR fixes this error in the docstring.


sanderland (approved):
Thank you for your contribution!

abdullahkady (approved):
thank you!

PR #342, opened 2023-11-14
poetry update, docs and workflows

sanderland:
Also incorporates the suggestion from #289


pyproject.toml
```diff
@@ -1,41 +1,38 @@
 [tool.poetry]
 name = "cohere"
-version = "4.34"
```
billytrend-cohere, line 3: should we major version if we're breaking ?

sanderland: It seems it's pretty standard to consider dropping support for EOL python versions to not be a breaking change (cf https://github.com/PyCQA/isort/releases -- I noticed it when upgrading black/isort/autoflake/... which have all done this) I agree it's not super clear though

lfayoux (approved):

PR #349, opened 2023-11-23
Update pre-commit hook

harry-cohere:
Updates one of the pre-commit hooks that we use to apply formatting, which solves an issue we were seeing on CI.

I ran this locally with `pre-commit run --all-files` and it succeeded.


harry-cohere:
I think we need https://github.com/cohere-ai/cohere-python/pull/349 to merge, as the commit hook that runs on dependency updates appears broken

billytrend-cohere (approved):

PR #354, opened 2023-12-05
Drop readthedocs and all references to it

billytrend-cohere:
readthedocs has been out of date for a long time so we shouldn't reference it and rather point to our main docs site which has code snippets etc.


harry-cohere (approved):

billytrend-cohere:
I guess in theory but it hasn't for a long time. Also I think we shouldn't make the docstrings canonical because they won't be as up to date as the docs.cohere.com documentation

billytrend-cohere:
Close for #355

PR #362, opened 2024-01-05
Add documentation for use of env vars for API key and URL

mjul:
Add doc strings for how the client uses environment variables as fallback for the API key and URL parameters.


lfayoux (approved):

lfayoux:
@mjul Thank you!

PR #373, opened 2024-01-29
Add frequency and presence penalty to Client.chat() and AsyncClient.chat()

liamdugan:
Added the `frequency_penalty` and `presence_penalty` parameters to `Client.chat()` and `AsyncClient.chat()`. 

These parameters are present for the `.generate()` method but are missing from the `.chat()` method despite being in the documentation (https://docs.cohere.com/reference/chat).


daniel-cohere (approved):
Thanks Liam!

PR #377, opened 2024-02-12
Add raw_prompting param to Generate

mkozakov:
Adds `raw_prompting` param to the Generate API. When set to True, the user's prompt will be sent to the model without any pre-processing.

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -162,6 +162,7 @@ class Client:
         truncate: Optional[str] = None,
         logit_bias: Dict[int, float] = {},
         stream: bool = False,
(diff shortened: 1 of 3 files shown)


cohere/client.py
```diff
@@ -162,6 +162,7 @@ def generate(
         truncate: Optional[str] = None,
         logit_bias: Dict[int, float] = {},
         stream: bool = False,
+        raw_prompting: bool = False,
```
lfayoux, line 165: also update async client?

lfayoux (approved):

PR #384, opened 2024-02-22
add finish reason to non stream chat

lfayoux:
add finish_reason to non stream chat


andyt-cohere:
context: eval team is migrating as many tasks as we can to Chat (we don't stream for our evaluations)

mkozakov (approved):

Issue #379, opened 2024-02-18
create_embed_job python SDK bug


khalid4294:
the create embed job function fails with error the input dataset_id is empty even though it's given. 
I fixed this error by going to:
> python3.9/site-packages/cohere/client.py
> Client
> create_embed_job

and I manually edited the json

from:

        json_body = {
            "input_dataset_id": input_dataset_id,
            "name": name,
            "model": model,
            "truncate": truncate,
            "text_field": text_field,
            "input_type": input_type,
            "output_format": "avro",
        }

to:

        json_body = {
            "input_dataset_id": input_dataset_id,
            "dataset_id": input_dataset_id,
            "name": name,
            "model": model,
            "truncate": truncate,
            "text_field": text_field,
            "input_type": input_type,
            "output_format": "avro",
        }

and the issue was fixed.

PR #387, opened 2024-02-25
FIX: wrong dataset_id param in create_embed_job

Mr-Destructive:
Fixing #379 
The `EmbedJob` constructor takes the field `dataset_id` as `input_dataset_id` which was mistakenly parsed as `database_id`

Thanks

### DIFF
--- a/cohere/client.py
+++ b/cohere/client.py
@@ -1055,7 +1055,7 @@ class Client:
         """
 
         json_body = {
-            "dataset_id": dataset_id,
+            "input_dataset_id": dataset_id,
             "name": name,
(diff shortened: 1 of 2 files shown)


lfayoux (commented):
The request param is `dataset_id` not `intput_dataset_id`

Mr-Destructive:
> The request param is `dataset_id` not `intput_dataset_id`

But here it is called as [input_dataset_id](https://github.com/cohere-ai/cohere-python/blob/main/cohere/responses/embed_job.py#L49) while parsing from the request to the actual object.

PR #408, opened 2024-03-19
Add local tokenizers

abdullahkady:


pyproject.toml
```diff
@@ -31,3 +35,9 @@ plugins = ["pydantic.mypy"]
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.extras]
+# used for local tokenization
+tokenizers = ["tokenizers", "requests"]
+# used for caching
+cachetools = ["cachetools"]
```
billytrend-cohere, line 43: what do these do is it config?

src/cohere/client.py
```diff
@@ -121,6 +124,19 @@ def __init__(
     delete_connector: Never = moved_function("delete_connector", ".connectors.delete")
     oauth_authorize_connector: Never = moved_function("oauth_authorize_connector", ".connectors.o_auth_authorize")
 
+    def local_tokenize(self, *, text: str, model: str) -> TokenizeResponse:
+        from .manually_maintained.tokenizers import local_tokenize
+        # TODO: parse into TokenizeResponse()? or should remain decoupled from the API responses?
```
billytrend-cohere, line 129: yeah i think so if it fits neatly into that type

src/cohere/manually_maintained/cache.py
```diff
@@ -0,0 +1,18 @@
...
+    # Optional caching mixin with time-to-live. If cachetools is not installed, caching is disabled.
+    cache = None
+    try:
+        from cachetools import TTLCache
+
+        cache = TTLCache(maxsize=100, ttl=60)
+    except ImportError:
+        pass
+
+    def cache_get(self, key):
```
billytrend-cohere, line 11: will these be exposed to user on the client object?

billytrend-cohere:
btw you can add tests to https://github.com/cohere-ai/cohere-python/tree/v5/tests and it will have access to the api key in ci

pyproject.toml
```diff
@@ -13,6 +13,10 @@ python = "^3.8"
 httpx = ">=0.21.2"
 pydantic = ">= 1.9.2"
 typing_extensions = ">= 4.0.0"
+tokenizers = {version="^0.15.2", optional=true}
+requests = {version="^2.26.0", optional=true}
+cachetools = {version="^5.3.3", optional=true}
```
billytrend-cohere, line 18: pr to generate these: https://github.com/cohere-ai/fern-cohere/pull/29

src/cohere/manually_maintained/cache.py
abdullahkady: good catch!

src/cohere/client.py
abdullahkady: After a deep dive .. the answer is no. 
In tokenize endpoint, we return the textual representation of **each** token, which is not achievable unless we called `decode()` on the encoded token ids. Since the local tokenization flow is supposed to optimize for speed, this might not seem like a good idea.

pyproject.toml
abdullahkady: yeah just like the concept of dependencies, they are not installed by default when you do `poetry install`

src/cohere/client.py
```diff
@@ -121,6 +124,19 @@ def __init__(
...
+    # def detokenize(
+    #     self,
+    #     *,
+    #     tokens: typing.Sequence[int],
+    #     model: typing.Optional[str] = OMIT,
+    # ) -> DetokenizeResponse:
+    #     return None
+
 
 class AsyncClient(AsyncBaseCohere):
```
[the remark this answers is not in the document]
abdullahkady: yeah good point!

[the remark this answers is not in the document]
abdullahkady: I'm still blocked because this depends on the API implementation of `GET /models/model-id` being in the SDK. Will think of a workaround for now.

src/cohere/client.py
```diff
@@ -4,6 +4,7 @@
 
 from .base_client import BaseCohere, AsyncBaseCohere
 from .environment import ClientEnvironment
+from .manually_maintained.cache import CacheMixin
```
abdullahkady, line 7: I'm hoping for a better name for `manually_maintained`. @billytrend-cohere any suggestions?

src/cohere/manually_maintained/tokenizers.py
```diff
@@ -0,0 +1,51 @@
...
+    if tokenizer is not None:
+        return tokenizer
+
+    # ---------- should use co.here API to get the tokenizer URL once the GET model is in place ----------
+    class FakeRes:
+        def json(self):
+            return {"tokenizer_url": "https://storage.googleapis.com/cohere-assets/tokenizers/command-v1.json"}
+
+    response = FakeRes()
+    # ---------- should use co.here API to get the tokenizer URL once the GET model is in place ----------
```
abdullahkady, line 33: Waiting to change base to the new generated SDK with the get model endpoint.

pyproject.toml
```diff
@@ -14,6 +14,8 @@ httpx = ">=0.21.2"
 pydantic = ">= 1.9.2"
 tokenizers = "^0.15.2"
 typing_extensions = ">= 4.0.0"
+# tokenizers = {version="^0.15.2", optional=true}
```
abdullahkady, line 17: revisit this before merging.

src/cohere/client.py
```diff
@@ -185,17 +197,107 @@ def embed(
...
-class AsyncClient(AsyncBaseCohere):
+        client = Client(
+            client_name="YOUR_CLIENT_NAME",
+            token="YOUR_TOKEN",
+        )
+        client.tokenize(
+            text="tokenize me! :D",
+            model="command",
+        )
+        """
```
billytrend-cohere, line 231: i think you can delete this because the docstring will get inheritted from the auto generated class

src/cohere/client.py
```diff
@@ -185,17 +197,107 @@ def embed(
...
+        # We cannot make the parameter required in the signature because otherwise the inheritance will break (mypy incompatible with supertype)
+        if model is OMIT:
+            raise ValueError("The model parameter is required.")
```
billytrend-cohere, line 234: can we fix this in the spec so we can fix the types?

src/cohere/client.py
abdullahkady: oh it does? :o there is a problem though with "appending to the docstring" for the `offline` param

src/cohere/client.py
abdullahkady: Yeah but it has to be done on the endpoint. I can do that now instead ✅

src/cohere/manually_maintained/tokenizers.py
```diff
@@ -0,0 +1,50 @@
...
+TOKENIZER_CACHE_KEY = "tokenizers"
+
+
+def tokenizer_cache_key(model: str) -> str:
+    return f"{TOKENIZER_CACHE_KEY}:{model}"
+
+
+async def get_hf_tokenizer(co: typing.Union["AsyncClient", "Client"], model: str) -> Tokenizer:
+    """Returns a HF tokenizer from a given tokenizer config URL."""
+    from cohere.client import AsyncClient
```
billytrend-cohere, line 17: why is this dynamically imported?

abdullahkady: to fix circular imports

src/cohere/manually_maintained/cache.py
```diff
@@ -0,0 +1,23 @@
+import typing
+import time
+
+
+class CacheMixin:
+    # A simple in-memory cache with TTL (thread safe). This is used to cache tokenizers at the moment.
+    _cache: typing.Dict[str, typing.Tuple[typing.Optional[float], typing.Any]] = dict()
```
billytrend-cohere, line 7: does it persist accross executions?

src/cohere/manually_maintained/tokenizers.py
abdullahkady: it's only used to check if it's an instance of Async/sync to avoid having two different methods

src/cohere/manually_maintained/cache.py
abdullahkady: no. it's per client (so if you define multiple clients in one runtime, each will contain its own cache)

src/cohere/client.py
billytrend-cohere: we could try this? https://stackoverflow.com/a/2025599

src/cohere/manually_maintained/cache.py
billytrend-cohere: ook, i'm thinking this might be slow for users that are running a script many times?

src/cohere/client.py
billytrend-cohere: looks fine to me, we can always revisit since it's not public

src/cohere/client.py
abdullahkady: wouldn't work because we need to append halfway through the parent doc for `parameters` section 😅
I'm okay completely removing it, and not having the `offline` param documented at all (I added it as an extra thing in case someone **needs** to hit the endpoint instead.

billytrend-cohere: yeah i think drop it for now so we can lean on the autogenerated one

src/cohere/manually_maintained/cache.py
abdullahkady: if they re init the client, yes. 
I was already looking into adding a persistent FS cache, but I thought that can come as an enhancement?

billytrend-cohere: yeah sg

src/cohere/client.py
abdullahkady: done

src/cohere/client.py
abdullahkady: done

PR #437, opened 2024-03-26
Add support for batching in .embed(...)

billytrend-cohere:
We manually override embed to allow seamless batching. This still uses autogenerated code but may require manually updating at times

### DIFF
--- a/src/cohere/client.py
+++ b/src/cohere/client.py
@@ -1,16 +1,20 @@
+import asyncio
 import typing
+from concurrent.futures import ThreadPoolExecutor
 
 import httpx
 
-from .base_client import BaseCohere, AsyncBaseCohere
+from . import EmbedResponse, EmbedInputType, EmbedRequestEmbeddingTypesItem, EmbedRequestTruncate
+from .base_client import BaseCohere, AsyncBaseCohere, OMIT
+from .config import embed_batch_size
+from .core import RequestOptions
 from .environment import ClientEnvironment
-from .utils import wait, async_wait
+from .utils import wait, async_wait, merge_embed_responses
 import os
 
 from .overrides import run_overrides
 
 run_overrides()
-
 
 # Use NoReturn as Never type for compatibility
 Never = typing.NoReturn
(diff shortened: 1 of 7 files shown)


src/cohere/client.py
```diff
@@ -97,6 +101,44 @@ def __exit__(self, exc_type, exc_value, traceback):
 
     wait = wait
 
+    _executor = ThreadPoolExecutor()
+
+    def embed(
```
mkozakov, line 106: Can we add a docstring so that help(co.embed) shows all the parameters with descriptions

src/cohere/utils.py
```diff
@@ -31,8 +32,8 @@ def get_validation_status(awaitable: typing.Union[EmbedJob, DatasetsGetResponse]
 
 def get_job(cohere: typing.Any,
             awaitable: typing.Union[CreateEmbedJobResponse, DatasetsCreateResponse, EmbedJob, DatasetsGetResponse]) -> \
-typing.Union[
-    EmbedJob, DatasetsGetResponse]:
+        typing.Union[
```
mkozakov, line 35: is this indentation correct?

src/cohere/utils.py
```diff
@@ -150,3 +151,81 @@ async def async_wait(
         raise Exception(get_failure_reason(job))
 
     return job
+
+
+def maybe_sum_field(obj: typing.Any, field: str) -> Optional[int]:
```
mkozakov, line 156: aggregate_truthy_field or something?

mkozakov (approved):

src/cohere/client.py
billytrend-cohere: fortunately the docs get inherited from the autogenerated base class so we can leave off here 
image

src/cohere/utils.py
billytrend-cohere: it is ok but not pretty, will follow up with better formatting config across the handwritten code

src/cohere/utils.py
billytrend-cohere: improved

PR #440, opened 2024-03-27
Fix bug when partial embeddings by type requested

billytrend-cohere:
Previously threw exception when not all embeddings type were requested

### DIFF
--- a/src/cohere/utils.py
+++ b/src/cohere/utils.py
@@ -211,13 +211,16 @@ def merge_embed_responses(responses: typing.List[EmbedResponse]) -> EmbedRespons
             for response in embeddings_type
         ]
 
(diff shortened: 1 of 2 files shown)


src/cohere/utils.py
```diff
@@ -211,13 +211,16 @@ def merge_embed_responses(responses: typing.List[EmbedResponse]) -> EmbedRespons
             for response in embeddings_type
         ]
 
+        # get non none fields from pydantic model
```
harry-cohere, line 214:
```suggestion
        # only get set keys from the pydantic model (i.e. exclude fields that are set to 'None')
```

harry-cohere (approved):

PR #452, opened 2024-04-03
Patch for pydantic 1.10.15

harry-cohere:


billytrend-cohere:
Fixes break caused by https://github.com/pydantic/pydantic/pull/9042

PR #482, opened 2024-04-19
Handle using the sync client in an async context for local tokenizers

abdullahkady:
Addressing this: https://github.com/langchain-ai/langchain-cohere/issues/19


harry-cohere:
We discussed and you'd like to try an alternative approach. Let me know when you'd like a review!

src/cohere/manually_maintained/tokenizers.py
```diff
@@ -12,39 +13,64 @@ def tokenizer_cache_key(model: str) -> str:
...
+    tokenizer = co._cache_get(tokenizer_cache_key(model))
+    if tokenizer is not None:
+        return tokenizer
+    tokenizer_url = co.models.get(model).tokenizer_url
+    if not tokenizer_url:
+        raise ValueError(f"No tokenizer URL found for model {model}")
+
+    size = int(typing.cast(int, requests.head(tokenizer_url).headers.get("Content-Length")))
+    size_mb = round(size / 1024 / 1024, 2)
+    print(f"Downloading tokenizer for model {model}. Size is {size_mb} MBs.")
```
harry-cohere, line 27: Did we decide on print versus logging? Logging is a better approach, but I can see that we don't do much of it within this SDK

harry-cohere (approved):
Looking good, thanks!

Would you be able to add integration tests?

src/cohere/manually_maintained/tokenizers.py
abdullahkady: I don't think we decided on anything. I saw that the SDK didn't have any logging setup, so I just went with print statements (since it's just progress indication for the user, not error/structured logs)

@billytrend-cohere do you have opinions on logging?

abdullahkady:
> Looking good, thanks!
> 
> Would you be able to add integration tests?

@harry-cohere I added a test but it doesn't make a lot of sense, now that we have split the functions (just like the rest of the SDK). 
The test would need to be something like:
- **async** test (in terms of the test context itself)
- initialize **sync** client
- run all the sync client calls (defined in the `test_client.py` in the test's async context)

src/cohere/manually_maintained/tokenizers.py
harry-cohere: I would recommend choosing logging over print, as people who use this library will likely want some level of control over how and when log lines are printed.

harry-cohere:
> > Looking good, thanks!
> > Would you be able to add integration tests?
> 
> @harry-cohere I added a test but it doesn't make a lot of sense, now that we have split the functions (just like the rest of the SDK). The test would need to be something like:
> 
> * **async** test (in terms of the test context itself)
> * initialize **sync** client
> * run all the sync client calls (defined in the `test_client.py` in the test's async context)

Perhaps just integration tests around the functionality, as it's intended to be used? Up to you though.

harry-cohere (approved):

PR #494, opened 2024-05-07
Return token_strings in Client.tokenize() when offline=True

### LINKED ISSUE #493: Offline tokenization produces empty token_strings
When I run the script on this doc: https://docs.cohere.com/reference/tokenize

```
response = co.tokenize(text="tokenize me! :D", model="command")
```

I get:

```
tokens=[10002, 2261, 2012, 8, 2792, 43] token_strings=[] meta=None
```

where `token_strings` is an empty array, even thought the docs suggests that it should be non-empty. However, if I run:

```
response = co.tokenize(text="tokenize me! :D", model="command", offline=False)
```

I get the `token_strings` as expected:

```
tokens=[10002, 2261, 2012, 8, 2792, 43] token_strings=['token', 'ize', ' me', '!', ' :', 'D'] meta=ApiMeta(api_version=ApiMetaApiVersion(version='1', is_deprecated=None, is_experimental=None), billed_units=None, tokens=None, warnings=None)
```

It would be nice if `token_strings` could be supported for offline tokenization, so that the online and offline behavior is identical. I'll attach a pull request for how this could be done.

yifanmai:
Fixes #493

### DIFF
--- a/src/cohere/client.py
+++ b/src/cohere/client.py
@@ -212,7 +212,8 @@ class Client(BaseCohere, CacheMixin):
         if offline:
             try:
                 tokens = local_tokenizers.local_tokenize(self, text=text, model=model)
-                return TokenizeResponse(tokens=tokens, token_strings=[])
+                token_strings = local_tokenizers.local_tokens_to_token_strings(self, tokens=tokens, model=model)
+                return TokenizeResponse(tokens=tokens, token_strings=token_strings)
             except Exception:
                 # Fallback to calling the API.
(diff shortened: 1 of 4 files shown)


abdullahkady:
Thanks @yifanmai for the contribution!

We intentionally skipped this when we introduced offline local tokenization; since we were optimizing for time. However, let me check how much overhead this adds (for bigger inputs) and if it's negligible we can add it.

yifanmai:
Noted, thanks!

PR #512, opened 2024-05-30
Add support for sagemaker

billytrend-cohere:
This PR introduces the `AwsClient` class and its implementation in the `aws_client.py` module. The `AwsClient` class is now extended by both the `BedrockClient` and `SagemakerClient` classes, which have been updated to accept additional parameters for AWS authentication and configuration.

- The `aws_client.py` module is added, which contains the `AwsClient` class and its methods.
- The `BedrockClient` class now inherits from `AwsClient` and accepts additional parameters for AWS authentication and configuration.
- The `SagemakerClient` class is introduced, inheriting from `AwsClient` and providing a client for SageMaker-specific functionality.
- The `embed_job.jsonl` file is imported in the `tests/test_aws_client.py` module but not used.
- The `test_aws_client.py` module is updated to include tests for both the `BedrockClient` and `SagemakerClient` classes, verifying their functionality with the new authentication parameters.


src/cohere/bedrock_client.py
```diff
@@ -33,229 +21,15 @@ def __init__(
...
-    def _event_hook(request: httpx.Request) -> None:
-        headers = request.headers.copy()
-        del headers["connection"]
-
-        endpoint = request.url.path.split("/")[-1]
-        body = json.loads(request.read())
-
-        url = get_url(
-            platform=service,
+            service="bedrock",
```
lusmoura, line 26: nit: should we have an enum for the service names?

lusmoura (approved):

src/cohere/bedrock_client.py
billytrend-cohere: oops i merged before i saw this but yeah I'll follow up

PR #533, opened 2024-06-26
Add tests for langchain

Anirudh31415926535:


billytrend-cohere:
Thanks for this @Anirudh31415926535, looks good but pls can we put it in a separate workflow in the file so that ci times don't start adding up

billytrend-cohere (approved):

.github/workflows/ci.yml
```diff
@@ -36,7 +36,53 @@ jobs:
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Set up python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.10.14
+      - name: Bootstrap poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python - -y --version 1.5.1
```
harry-cohere, line 51: Can we use a pinned version of https://github.com/marketplace/actions/install-poetry-action instead?

.github/workflows/ci.yml
```diff
@@ -36,7 +36,53 @@ jobs:
...
+          poetry install --with test 
+
+          echo "Going to uninstall cohere package, and install the current version from the repo"
+          cd ../../../cohere-python && echo "Current dir $(pwd)"
+          echo "Current dir $(ls)"
+          pip uninstall cohere
+          pip install .
+          cd ../langchain-cohere/libs/cohere
+          
+          echo "Current cohere installation: $(pip freeze | grep cohere)"
```
harry-cohere, line 72: Might be a wise idea to echo the location too, something like:

`python -c 'import cohere ; print(cohere.__file__)'`

.github/workflows/ci.yml
```diff
@@ -36,7 +36,53 @@ jobs:
...
+          make test
```
harry-cohere, line 73: [As per the Makefile ](https://github.com/langchain-ai/langchain-cohere/blob/main/libs/cohere/Makefile), the package separates unit tests vs integration/replay tests, with `make test` only running the former
```suggestion
          make test
          make integration_test
```

.github/workflows/ci.yml
harry-cohere: If we did this we'd be able to clearly see that it's using the local version. With pip freeze we'd only get to see the package version number, which might confuse us into thinking it's the local version when it's not.

Anirudh31415926535: Actually we can see the file path via pip freeze since we installed thorugh the file path: 
`
Current cohere installation: cohere @ file:///home/runner/work/cohere-python/cohere-python`

Issue #534, opened 2024-07-03
Allow users to configure `embed_batch_size` or `ThreadPoolExecutor` size when calling `Client.embed`


acompa:
It looks like batching was added in #437 - thank you for implementing this, it's very helpful.

I notice that batching, as defined [here](https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/client.py#L135), depends on a [fixed batch size](https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/config.py#L1). This can be suboptimal for clients submitting a large number of smaller documents, as we cannot [configure the ThreadPoolExecutor size](https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/client.py#L111) to parallelize a large number of small data payloads. As a result a client might end up blocking while waiting for small network responses.

Would it be possible to allow clients to configure either the `ThreadPoolExecutor` size or the `embed_batch_size` setting when calling `embed`?

billytrend-cohere:
Hey @acompa thanks for the feedback, this will be fixed by https://github.com/cohere-ai/cohere-python/pull/536 ! you will be able to pass your own executor in. I'll ping the thread when it's released

PR #536, opened 2024-07-09
Add thread_pool_executor to constructor

billytrend-cohere:
The PR introduces changes to the `Client` and `AsyncClient` classes in `src/cohere/client.py`. 

## Summary
The changes add a `_executor` attribute to both classes and set its value during initialization. The `ThreadPoolExecutor` is now passed as an argument during initialization, allowing for more flexibility and customization. 

## Changes
- The `_executor` attribute is added to the `Client` and `AsyncClient` classes.
- During initialization, the `thread_pool_executor` argument is added to the `__init__` method of both classes.
- The value of `self._executor` is set to the provided `thread_pool_executor` argument.
- The previous instantiation of `_executor` outside the methods is removed.

### DIFF
--- a/src/cohere/client.py
+++ b/src/cohere/client.py
@@ -71,6 +71,8 @@ def deprecated_function(fn_name: str) -> typing.Any:
 
 
 class Client(BaseCohere, CacheMixin):
+    _executor: ThreadPoolExecutor
+
     def __init__(
         self,
         api_key: typing.Optional[typing.Union[str, typing.Callable[[], str]]] = None,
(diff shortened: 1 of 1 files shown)


fede-dash:
Hi\! I wanted to follow up on this PR as I've just submitted PR #699 which complements this implementation.

This PR addressed part of issue #534 by allowing ThreadPoolExecutor configuration at the client level, which is great. However, issue #534 also mentioned the need for configurable batch_size, and some users might need per-call flexibility rather than client-wide settings.

PR #699 adds:
- `batch_size` parameter to the embed() method (the other key request from #534)
- `max_workers` parameter for per-call thread pool override
- Dynamic optimization without recreating the client

Both PRs work together nicely - users can set a default ThreadPoolExecutor via your implementation and then override it for specific operations using the per-call parameters.

Just wanted to connect these implementations since they're solving related parts of the same issue\!

Issue #559, opened 2024-08-10
Pydantic V2 warning


n1lanjan:
**SDK Version (required)**
Provide the version you are using.
```console
❯ python
Python 3.11.4 (main, Feb  2 2024, 19:01:43) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cohere
/user/home/.venv/lib/python3.11/site-packages/pydantic/_internal/_config.py:341: UserWarning: Valid config keys have changed in V2:
* 'allow_population_by_field_name' has been renamed to 'populate_by_name'
* 'smart_union' has been removed
  warnings.warn(message, UserWarning)
>>> print(cohere.__version__)
5.8.0
>>> import pydantic
>>> pydantic.VERSION
'2.8.2'
```

**Describe the bug**

I am getting this warning while trying to use the library.
```
.venv/lib/python3.11/site-packages/pydantic/_internal/_config.py:341: UserWarning: Valid config keys have changed in V2:
* 'allow_population_by_field_name' has been renamed to 'populate_by_name'
* 'smart_union' has been removed
  warnings.warn(message, UserWarning)
```

This warning is triggering from the code below:

- https://github.com/cohere-ai/cohere-python/blob/540207821d15fcf933935e44f76a07cd69e53465/src/cohere/core/pydantic_utilities.py#L85-L90

I proposed a PR, please review
- https://github.com/cohere-ai/cohere-python/pull/560

billytrend-cohere:
Thanks @n1lanjan we are looking into this!

billytrend-cohere:
This is now fixed in the latest version of the sdk! thanks @n1lanjan

Tenzer:
The latest version still triggers deprecation warnings with Pydantic V2. You can see this if you import the Cohere library while elevating warnings to errors (like we do in our test suite):

```bash
$ python -W all -c 'import cohere'
<path>/.venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:273: PydanticDeprecatedSince20: `json_encoders` is deprecated. See https://docs.pydantic.dev/2.8/concepts/serialization/#custom-serializers for alternatives. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
  warnings.warn(
[...]

PR #560, opened 2024-08-10
Update pydantic_utilities.py

n1lanjan:
- https://github.com/cohere-ai/cohere-python/issues/559


dsinghvi (approved):

billytrend-cohere:
Thanks for this suggestion, @n1lanjan we have worked with our sdk vendors to upstream your fix and the latest version of our sdk fixes this!

PR #588, opened 2024-09-30
Sagemaker finetuning

billytrend-cohere:
Generating pr description


collindutter:
Hey @billytrend-cohere! The introduction of `sagemaker` has introduced some dependency resolution issues in our [library](https://github.com/griptape-ai/griptape). Any possibility of making this an optional dependency? Otherwise we will need to pin `5.10.0`. Thanks!

PR #616, opened 2024-12-04
Throw when trying to rerank with v1

billytrend-cohere:


src/cohere/bedrock_client.py
```diff
@@ -25,6 +25,8 @@ def __init__(
             timeout=timeout,
         )
 
+    def rerank(self, *, query, documents, model = ..., top_n = ..., rank_fields = ..., return_documents = ..., max_chunks_per_doc = ..., request_options = None):
+        raise NotImplementedError("Please use cohere.BedrockClientV2 instead: method rerank is not available in the v1 client, cohere.BedrockClient.")
```
mkozakov, line 29: this makes it sound like we haven't added support in the V1 sdk yet. should we change the message t o say
```suggestion
        raise NotImplementedError("Please use cohere.BedrockClientV2 instead: Rerank API on bedrock is not supported with cohere.BedrockClient version 1 for this model.")
```

mkozakov (approved):

PR #618, opened 2024-12-05
bump numpy

mkozakov:
This PR updates the Python version from 3.8 to 3.9 and adds the `numpy` package as an optional dependency.

- **Python version update**: The Python version has been updated from `^3.8` to `^3.9`, ensuring compatibility with the latest features and improvements.
- **Addition of numpy package**: The `numpy` package is introduced as an optional dependency, providing powerful numerical computing capabilities for data manipulation and analysis.


andrewbcohere (approved):

andrewbcohere (approved):

Tenzer:
I fail to see the point of this change.

cohere-python library isn't using numpy, from what I can tell, and this change limits the max numpy version consumers of cohere-python can have to be less than version 2 (the latest numpy version is 2.1.3).

Issue #711, opened 2025-12-18
`image_tokens` missing from embedding response model


nathan-chappell:
**SDK Version (required)**
5.20.0

**Describe the bug**
Type for response from embedding API does not match with response.

I don't know if you guys need to regenerate the SDK from the API, update the API definition, or find a new generator, but the `ApiMetaBilledUnits` has an attribute `images`, while the response returned has an attribute `image_tokens`.  

The generator using `# type: ignore` and `ConfigDict(extra='allow')` hides the issue.

```
# printing the billed units from a response
print(resp.meta.billed_units.model_dump_json(indent=2))
# output
{
  "images": null,
  "input_tokens": 0.0,
  "output_tokens": null,
  "search_units": null,
  "classifications": null,
  "image_tokens": 2716 # this attribute is not featured in the model
}
```

PR #784, opened 2026-07-26
fix(utils): batched embed drops meta.tokens and image billed units

arthi-arumugam-git:
`Client.embed()` goes through `merge_embed_responses` on every call unless you pass `batching=False` or pass images, and that path rebuilds `ApiMeta` from scratch inside `merge_meta_field`. The rebuild sets `api_version`, four of the six `ApiMetaBilledUnits` fields, and `warnings`. Everything else on the metadata is dropped on the floor: `meta.tokens`, `meta.cached_tokens`, `billed_units.images` and `billed_units.image_tokens`.

Since `batching` defaults to `True`, the default path is the lossy one:

```python
co.embed(texts=texts).meta.tokens                  # None
co.embed(texts=texts, batching=False).meta.tokens  # populated
```

Same request, different metadata, and nothing in the signature suggests that a batching flag should change which usage numbers come back. Anyone reading `meta.tokens` for accounting gets `None` and no error to tell them why.

The fix sums the four missing fields exactly the way the existing four are summed. `cached_tokens` sits directly on `ApiMeta` rather than under `billed_units`, so it sums off the meta list instead of the billed units list.

One deliberate choice worth flagging: `tokens` stays `None` when none of the input metas carried it, rather than building an `ApiMetaTokens(None, None)`. A merged response with no token counts then looks exactly like it does today, which is why the existing equality assertions in `test_embed_utils.py` did not need touching.

### Why this happened, and stopping it happening again

`merge_meta_field` lists the fields it copies by hand. `images`, `image_tokens`, `tokens` and `cached_tokens` were added to the models and the merge was never updated, and nothing failed. The same thing will happen to the next field added.

So one of the three tests drives its assertions off the model fields rather than a hardcoded list:

```python
billed_fields = get_fields(ApiMetaBilledUnits())
...
for field in billed_fields:
    self.assertEqual(getattr(merged.billed_units, field), 2, f"billed_units.{field} was dropped")
```

Add a field to `ApiMetaBilledUnits` or `ApiMetaTokens` without touching `merge_meta_field` and that test fails immediately. I verified it fails on `main` as it stands.

The other two are ordinary: `test_merge_meta_field_keeps_tokens_and_image_units` covers the summing and uses `output_tokens=0` on both inputs so a falsy value is not confused with an absent one, and `test_merge_meta_field_leaves_tokens_unset_when_absent` pins the `None` behaviour so it does not get simplified away later.

### Notes

- Only `src/cohere/utils.py` and `tests/` are touched, both listed in `.fernignore`.
- `mypy .` clean across 341 files, `tests/test_embed_utils.py` 9 passed.
- I originally included a `7.0.8 -> 7.0.9` bump following #778 and #779 and have dropped it. I have a second fix open against `utils.py` at the same time and they cannot both bump to the same version, and `core/client_wrapper.py` is Fern generated rather than hand maintained, so the version looks like yours to set at release. Say the word if you would rather the bump were in here and I will add it back.

---

> [!NOTE]
> **Low Risk**
> Localized change to embed response merging and tests; fixes incorrect metadata rather than altering auth or API contracts.
> 
> **Overview**
> **Fixes silent loss of usage metadata** when `Client.embed()` merges multiple batch responses (default `batching=True`). `merge_meta_field` now sums **`billed_units.images`**, **`billed_units.image_tokens`**, **`meta.tokens`** (input/output), and **`cached_tokens`** alongside the fields it already merged, so batched calls return the same usage shape as a single request.
> 
> **`tokens` stays `None`** when no input meta had token counts (no empty `ApiMetaTokens`), keeping merged responses aligned with unbatched ones when token data was never present.
> 
> **Tests** cover image/token summing, absent-token behavior, and a model-driven check that every numeric field on `ApiMetaBilledUnits` / `ApiMetaTokens` is merged so future model fields cannot be dropped without failing CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0878e27d22e2fd7c996ab9e0710d9567fe48605f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cohere/utils.py
+++ b/src/cohere/utils.py
@@ -9,7 +9,7 @@ import requests
 from fastavro import parse_schema, reader, writer
 
 from . import EmbedResponse, EmbeddingsFloatsEmbedResponse, EmbeddingsByTypeEmbedResponse, ApiMeta, \
-    EmbedByTypeResponseEmbeddings, ApiMetaBilledUnits, EmbedJob, CreateEmbedJobResponse, Dataset
+    EmbedByTypeResponseEmbeddings, ApiMetaBilledUnits, ApiMetaTokens, EmbedJob, CreateEmbedJobResponse, Dataset
 from .datasets import DatasetsCreateResponse, DatasetsGetResponse
 from .overrides import get_fields
 
@@ -170,19 +170,37 @@ def sum_fields_if_not_none(obj: typing.Any, field: str) -> Optional[int]:
 def merge_meta_field(metas: typing.List[ApiMeta]) -> ApiMeta:
     api_version = metas[0].api_version if metas else None
     billed_units = [meta.billed_units for meta in metas]
+    images = sum_fields_if_not_none(billed_units, "images")
     input_tokens = sum_fields_if_not_none(billed_units, "input_tokens")
+    image_tokens = sum_fields_if_not_none(billed_units, "image_tokens")
     output_tokens = sum_fields_if_not_none(billed_units, "output_tokens")
     search_units = sum_fields_if_not_none(billed_units, "search_units")
     classifications = sum_fields_if_not_none(billed_units, "classifications")
+
+    token_counts = [meta.tokens for meta in metas]
+    token_input = sum_fields_if_not_none(token_counts, "input_tokens")
+    token_output = sum_fields_if_not_none(token_counts, "output_tokens")
+    # Leave tokens unset rather than building an all-None ApiMetaTokens, so a
+    # merged response is indistinguishable from an unbatched one.
+    tokens = ApiMetaTokens(
+        input_tokens=token_input,
+        output_tokens=token_output,
+    ) if token_input is not None or token_output is not None else None
+
+    cached_tokens = sum_fields_if_not_none(metas, "cached_tokens")
     warnings = {warning for meta in metas if meta.warnings for warning in meta.warnings}
     return ApiMeta(
         api_version=api_version,
         billed_units=ApiMetaBilledUnits(
+            images=images,
             input_tokens=input_tokens,
+            image_tokens=image_tokens,
             output_tokens=output_tokens,
             search_units=search_units,
             classifications=classifications
         ),
+        tokens=tokens,
+        cached_tokens=cached_tokens,
         warnings=list(warnings)
     )
 
--- a/tests/test_embed_utils.py
+++ b/tests/test_embed_utils.py
@@ -1,8 +1,9 @@
 import unittest
 
 from cohere import EmbeddingsByTypeEmbedResponse, EmbedByTypeResponseEmbeddings, ApiMeta, ApiMetaBilledUnits, \
-    ApiMetaApiVersion, EmbeddingsFloatsEmbedResponse
-from cohere.utils import merge_embed_responses, sum_fields_if_not_none
+    ApiMetaApiVersion, ApiMetaTokens, EmbeddingsFloatsEmbedResponse
+from cohere.overrides import get_fields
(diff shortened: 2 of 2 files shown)


arthi-arumugam-git:
Good catch from the bot, the User-Agent string was still on 7.0.8 while X-Fern-SDK-Version and pyproject.toml were on 7.0.9. Pushed a fix so both read cohere/7.0.9.

Note that #785 also bumps 7.0.8 -> 7.0.9, so whichever of the two lands second will need its bump rebased. Happy to rebase on request, or drop the bump commit from one of them if you would rather handle versioning at release time.

arthi-arumugam-git:
Pushed a cleanup pass on both this and #785.

Dropped the version bump from both. They cannot both go to 7.0.9, and `core/client_wrapper.py` is Fern generated rather than in `.fernignore`, so it is not really mine to edit. That also resolves the User-Agent mismatch Bugbot flagged, since the bump is gone. Each PR now touches only `src/cohere/utils.py` and `tests/`, so they no longer conflict with each other and can merge in either order.

Also added a test here that drives its assertions off the model fields instead of a hardcoded list. `merge_meta_field` enumerates what it copies by hand, which is exactly why these four fields went missing when they were added, so the new test fails the moment a field is added to `ApiMeta` or `ApiMetaBilledUnits` without updating the merge. Verified it fails on `main` as it stands. Body updated with the details.

arthi-arumugam-git:
Some prior art I found after opening this, which is worth linking.

Issue #711 reported `image_tokens` missing from the embedding response model and was closed on 2026-01-16 when the model gained the field through a regeneration. But #711 had two halves. The model was one, and `merge_meta_field` copying it through was the other, and only the first landed. So on `main` today `ApiMetaBilledUnits.image_tokens` exists and is populated on an unbatched call, and is still dropped on a batched one. #711 reads as fixed and is not.

PR #714 spotted that and has been open since 2025-12-28. Its model change was overtaken by the regeneration; its one line adding `image_tokens` to `merge_meta_field` never merged. This PR includes that line, alongside `images`, `tokens` and `cached_tokens`, so if this lands then #714 can be closed. Credit to @yurekami for finding it first, seven months ago.

That is also the argument for the model-driven test rather than another hardcoded assertion. A field was added to the model, the merge was not updated, an issue was filed, the issue was closed, and the bug survived all of it because nothing failed. The test I added fails in that situation.

arthi-arumugam-git:
Bugbot reviewed afab009, which is no longer in this branch. The rebase at 0878e27 drops the version bump and the `client_wrapper.py` change entirely, so the User-Agent mismatch it flagged is gone.

Reason for dropping it rather than fixing it: two open PRs cannot both claim 7.0.9, and `src/cohere/core/client_wrapper.py` is Fern generated rather than hand maintained. Versioning is yours to do at release time.

This PR now touches `src/cohere/utils.py` and `tests/test_embed_utils.py` only, both covered by `.fernignore`.

arthi-arumugam-git:
A short follow-up while this sits, in case it helps whoever picks it up.

When embed calls are batched, the per-batch responses are merged but `meta.billed_units.tokens` and the image billed units are not carried into the merged result, so they come back absent. Anyone reconciling their own usage against what Cohere billed sees the tokens for a single-batch call and nothing for a batched one, which reads as a drop in usage rather than a reporting gap.

`cohere-python` is largely Fern generated, so I kept the change inside the hand-written merge helper in `utils.py` rather than anywhere the generator would overwrite it. If that is the wrong place for it, or if this belongs upstream in the Fern template instead, I am happy to move it.

